PR #2, opened 2015-07-23
Add DebugEvent

janvernieuwe:
Has the output of debugLastSoapRequest as third parameter


veewee:
Debug is not an event. You could create a listener for accessing debug information on every response / fault event.

Issue #4, opened 2016-04-20
Patching does not add new elements


harmonwood:
When adding a new element to a WSDL node and then run the generate:types command again it does not get added as a class attribute via an `<xsd:import schemaLocation ..../>` file.
WSDL was:

    ....
    <xs:complexType name="Field">
        <xs:sequence>
            <xs:element minOccurs="0" name="Description" nillable="true" type="xs:string"/>
            <xs:element minOccurs="0" name="FieldType" nillable="true" type="xs:string"/>
        </xs:sequence>
    </xs:complexType>
    ....

    ....
    <xs:complexType name="Field">
        <xs:sequence>
            <xs:element minOccurs="0" name="Description" nillable="true" type="xs:string"/>
            <xs:element minOccurs="0" name="FieldType" nillable="true" type="xs:string"/>
            <xs:element minOccurs="0" name="SortOrder" type="xs:int"/>
        </xs:sequence>
    </xs:complexType>
    ....

 It does create a backup file and state that it is patching like so:

    Type Field exists. Trying to patch ...Patched!

The file and the backup are identical without the new element in the class.

    class Field
    {

       /**
        * @var string
        */
       protected $Description;

       /**
        * @var string
        */
       protected $FieldType;

    }

harmonwood:
The patch file generated is:

    +++ /Field.php▸ 2016-04-20 10:24:44.000000000 -0500
    @@ -46,11 +46,6 @@
         protected $FieldType;
    -
    -    /**
    -     * @var int
    -     */
    -    protected $SortOrder;

veewee:
Hello @harmonwood,

Thanks for reporting the bug. I've added some failing tests but need to investigate how we van solve this issue. I'll keep you posted when there is any progess on the issue!

veewee:
Hi @harmonwood , @Aluok, @janvernieuwe,


I started working on the zend-code featue and got a first POC version going for the type generator:
https://github.com/veewee/soap-client/commit/00da7f41d32a65cf9fdd906da2e86288d200e19f

(note that the class generator is not implemented yet)

These files can be tested with for example:
```
./bin/soap-client generate:types --wsdl="http://wsf.cdyne.com/WeatherWS/Weather.asmx?WSDL" --namespace="Testing" gentest
```

The code works for both existing and new files and patches everything the way it should.
There are still some problems with zend-code like e.g:

- I cant remove a property from a class to overwrite it
- It uses full 'implements' instead of the short link

[...]

PR #7, opened 2016-06-19
Bug patching new elements

Aluok:
This is a solution to issue #4. No features added. 

I assumed that the tests with tampered fixture meant that you wanted to keep any functions that are defined, even getters/setters. Is that right?
The way it's currently done might need some input to be improved (in a sense, I don't really like the way it's done, but I don't see another...

Thanks for the project.

### DIFF
--- a/src/Phpro/SoapClient/CodeGenerator/Patcher.php
+++ b/src/Phpro/SoapClient/CodeGenerator/Patcher.php
@@ -38,29 +38,58 @@ class Patcher
         $original = realpath($original);
         $basename = pathinfo($original, PATHINFO_BASENAME);
         $tmpFile = $this->tmpFolder . DIRECTORY_SEPARATOR . time() . $basename;
+        $tmpFilePatched = $tmpFile . 'patch';
+        $functions = $this->getFunctions($original);
         $patchFile = $tmpFile . '.patch';
         $this->filesystem->putFileContents($tmpFile, $newContent);
-
+        $this->filesystem->putFileContents($tmpFilePatched, $this->filesystem->getFileContents($original));
         // Patch content:
         try {
             // Try to patch the original file into the new temporary file:
-            $patchData = $this->createPatch($original, $tmpFile, $patchFile);
+            $patchData = $this->createPatch($tmpFilePatched, $tmpFile, $patchFile);
             // No patch needs to be applied ...
             if (!$patchData) {
-                $this->cleanTmpFiles([$tmpFile, $patchFile]);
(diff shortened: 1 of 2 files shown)


src/Phpro/SoapClient/CodeGenerator/Patcher.php
```diff
@@ -38,29 +38,58 @@ public function patch($original, $newContent)
...
+        $this->filesystem->addToFile($original, $functions, $endClassPos);
+        $this->cleanTmpFiles([$tmpFile, $patchFile, $tmpFilePatched]);
+    }
+
+    /**
+     * @param string $original
+     *
+     * @return string
+     */
+    protected function getFunctions($original)
```
veewee: This method might need some tweeking: what about private, protected and functions without the public keyword?
Does it also work for methods without docblocks?
This functionality definetly needs some tests.

src/Phpro/SoapClient/Util/Filesystem.php
```diff
@@ -78,4 +78,22 @@ public function replaceFile($file, $new, $backup = true)
 
         $this->putFileContents($file, $this->getFileContents($new));
     }
+
+    /**
+     * @param string $file
+     * @param string $newContent
+     * @param int $pos the position at which to insert the content. (default to NULL, understood as append)
+     */
+    public function addToFile($file, $newContent, $pos = null)
```
veewee: The $pos is added by the implementation of this method, shouldn't it be a required parameter?

veewee:
Hi @Aluok,

The tests with tampered fixtures indeed meant that I wanted to keep any function that is not available in the new class. The downside of this is that it won't be able to remove properties that don't exist in the WSDL anymore.

The code looks pretty good but doesn't cover every possible case. I've added some small remarks to the code. We definitely need some more unit tests to make sure every works as expected.

In a next phase, I am planning to add [zend-code](http://framework.zend.com/manual/current/en/modules/zend.code.generator.introduction.html) instead of current patch implementation. This will make it possible to write your own code generation rules to speed up your development process even more.
Example rules could be: If a WSDL object name ends with `Request` -> Implement the `RequestInterface`. 

Anyway, this sure is a wanted feature. Thanks for your work! :)

veewee:
Hello @Aluok,

Since the new code generation commands are almost ready, I don't think it is a good idea to put any more time in this PR.
Feel free to check out #10 and give your remarks.

Thanks for the PR anyway!

Aluok:
Hello @veewee,

No problem. I didn't had much time, for the last days, to work more on it anyway. I'll try to give a look at #10 as soon as I can.
Thank you for your work on this project.

PR #12, opened 2016-07-28
set soap options in Console commands

malles:
This was preventing this function from working from my cli.

### DIFF
--- a/src/Phpro/SoapClient/Console/Command/GenerateClassmapCommand.php
+++ b/src/Phpro/SoapClient/Console/Command/GenerateClassmapCommand.php
@@ -73,7 +73,13 @@ class GenerateClassmapCommand extends Command
             throw InvalidArgumentException::invalidConfigFile();
         }
 
-        $soapClient = new SoapClient($config->getWsdl(), []);
+        $soapClient = new SoapClient($config->getWsdl(), [
+			'trace' => true,
+			'exceptions' => true,
+			'keep_alive' => true,
+			'cache_wsdl' => WSDL_CACHE_BOTH,
+			'features' => SOAP_SINGLE_ELEMENT_ARRAYS
+		]);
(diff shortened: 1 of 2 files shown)


veewee:
Hi @malles,

Thanks for the PR.
Can you give me some more information about which option solved your problem and what went wrong exactly?

You've added:
```
'trace' => true,
'exceptions' => true,
'keep_alive' => true,
'cache_wsdl' => WSDL_CACHE_BOTH,
'features' => SOAP_SINGLE_ELEMENT_ARRAYS
```

The trace is not needed here since we're not doing any debugging in the command. 
The exceptions part is a good addition. 
I don't think the keep-alive is necessary since there is no real SOAP request going on?
The cache is something you want to disable during develoment: you always want the latest WSDL.
The single elements looks something application specific?

Maybe it's a better idea to move the soap options to the `ConfigInterface`.
This way we can set some sensible defaults in the `Config` class and you can tune the configuration for your applications needs?

malles:
To be quite honest I do not know exactly what setting was needed to make it work. I just copied the same hardcoded options that are used in the Client.
It would make sense to put the config in a central place indeed.
Love the way the assemblers work! I just can't seem to find out how to add Custom Assemblers from my own namespace when working from the CLI.
My implementation is here: https://github.com/Bixie/pagekit-perfectview_api

veewee:
Hi @malles,

I've updated the code as mentioned above. (See PR #13)
Glad you like the assemblers, we've put a lot of effort in that section of the code :)

About adding assemblers from your own namespace:

- You'll have to add an autoloader section to your composer.json file and run composer install again.
- Next you'll need to add the correct namespace for the custom assembler in the configuration file. Like this: 

```
Use Bixie\PerfectviewApi\CodeGenerator\Assembler as BixieAssembler;

return Config::create()
	->addRule(new Rules\AssembleRule(new BixieAssembler\FluentSetterAssembler()))
```

Feel free to create a PR on the FluentSetterAssembler!

PR #14, opened 2016-07-30
Add FluentSetterAssembler

malles:
Creates fluent setters.

### DIFF
--- a/src/Phpro/SoapClient/CodeGenerator/Assembler/FluentSetterAssembler.php
+++ b/src/Phpro/SoapClient/CodeGenerator/Assembler/FluentSetterAssembler.php
@@ -41,7 +41,7 @@ class FluentSetterAssembler implements AssemblerInterface
                     'name' => $methodName,
                     'parameters' => [$property->getName()],
                     'visibility' => MethodGenerator::VISIBILITY_PUBLIC,
-                    'body' => sprintf('$this->%1$s = $%1$s;%2$sreturn $this;', $property->getName(), "\n"),
+                    'body' => sprintf('$this->%1$s = $%1$s;%2$sreturn $this;', $property->getName(), $class::LINE_FEED),
                     'docblock' => DocBlockGenerator::fromArray([
                         'tags' => [
                             [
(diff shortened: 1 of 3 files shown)


src/Phpro/SoapClient/CodeGenerator/Assembler/FluentSetterAssembler.php
```diff
@@ -0,0 +1,64 @@
...
+		$property = $context->getProperty();
+		try {
+			$methodName = Normalizer::generatePropertyMethod('set', $property->getName());
+			$class->removeMethod($methodName);
+			$class->addMethodFromGenerator(
+				MethodGenerator::fromArray([
+					'name' => $methodName,
+					'parameters' => [$property->getName()],
+					'visibility' => MethodGenerator::VISIBILITY_PUBLIC,
+					'body' => sprintf('$this->%1$s = $%1$s;%2$sreturn $this;', $property->getName(), "\n"),
```
malles: I'm not sure about the "\n" line-end. `$file` is not available int the TypeContext.

veewee: The LINE_FEED constant is located on the abstract zend generators. You can also use `$class::LINE_FEED`.

veewee:
Thanks for the PR, it looks pretty good!
Can you resolve the Travis issues? It looks like you've used tabs instead of 4 spaces.
Before I can merge this PR, we'll also need a phpunit test for the class and some documentation about the generator so that it can be used by everybody.

test/PhproTest/SoapClient/Unit/CodeGenerator/Assembler/FluentSetterAssemblerTest.php
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * @return TypeContext
+     */
+    private function createContext()
+    {
+        $class = new ClassGenerator('MyType', 'MyNamespace');
+        $type = new Type('MyNamespace', 'MyType', [
+            'prop1' => 'string',
+            'prop2' => 'int'
```
veewee: Maybe this prop2 should be removed since it isn't used?

veewee:
One small remark left. It's almost ready to merge :)

veewee:
Thanks!

PR #16, opened 2016-08-23
Fix logging prefix

GaryJones:


veewee:
Thanks for the PR!

GaryJones:
Thanks for a very useful package :-)

Issue #34, opened 2017-02-17
Ignoring null values when XML is created?


HKandulla:
Hi,

is it possible to make the generated xml not include any properties that are null?
Can I create a TypeConverter to accomplish this?

Thanks, Hannes

veewee:
Can you provide some additional information about what you are trying to do?
Maybe a code sample would be handy :)

HKandulla:
Sure. The Service I send my requests to allows me to send:
1. <ns1:DoNotCall>true</ns1:DoNotCall>
2. leave it out completely. 
But I am not allowed to send a null value in the form:
3. <ns1:DoNotCall xsi:nil="true“/>

veewee:
This can be done with a type converter, but then you'll have to add a converter for every type you want to convert this way. In this case: ns1:DoNotCall.

HKandulla:
first of all, tahnks for your quick response!
I created the type converter and used and defined:

`

    public function convertPhpToXml($php)
    {
        return !is_null($php)
            ?
            sprintf('<%1$s>%2$s</%1$s>', $this->getTypeName(), $php)
            :
            ''
            ;
    }
`

But the generated XML still contains `<ns1:CustomerID/>`

Any idea how I can eliminate the creation of the tag completely?

Thanks, Hannes

faibl:
Hi @veewee,

I figured that the type converter only allows to alter the content and not the tag itself. But I solved it by creating a middleware that removes all empty tags:

`

    public function beforeRequest(callable $handler, RequestInterface $request, array $options)
    {
        $xml = SoapXml::fromStream($request->getBody());

        // remove all empty nodes
        while (($notNodes = $xml->xpath('//*[not(node())]')) && ($notNodes->length)) {
            foreach($notNodes as $node) {
                $node->parentNode->removeChild($node);
            }
        }

        return $handler(new Request(
            $request->getMethod(),
            $request->getUri(),
            $request->getHeaders(),
            $xml->toStream()
        ), $options);
    }
`

Thanks for your help and a great library.

Best, Hannes

veewee:
No problem! Do you mind adding the Middleware to the package? I am pretty sure it will come in handy for other people as well. Since you already have the code, you'll only have to add some tests and some documentation.

Also, this might be a bit shorter then creating a new request manually:

```php
$request = $request->withBody($xml->toStream());
```

Thanks!

faibl:
sure, I am happy to contribute!
I prepare a PR later tonight.
[...]

PR #38, opened 2017-03-03
add decimal typehints and converter

HKandulla:


src/Phpro/SoapClient/Soap/TypeConverter/DecimalTypeConverter.php
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function convertXmlToPhp($data)
+    {
+        $doc = new DOMDocument();
+        $doc->loadXML($data);
+
+        if ('' === $doc->textContent) {
+            return null;
```
veewee: Should this method really return null? I would think that if it is null, it would return the value `0.0`. Otherwise we won't be able to add strict types to the generated code in the future. Or do you see use-cases for `null` values?

test/PhproTest/SoapClient/Unit/TypeConverter/DecimalTypeConverterTest.php
```diff
@@ -0,0 +1,81 @@
+<?php
+
+namespace PhproTest\SoapClient\Unit\Middleware;
```
veewee: It looks like you've placed this test in a wrong namespace.

test/PhproTest/SoapClient/Unit/TypeConverter/DecimalTypeConverterTest.php
```diff
@@ -0,0 +1,81 @@
...
+     */
+    public function testConvertPhpToXml()
+    {
+        $xml = '<decimal>24.7</decimal>';
+
+        $output = $this->converter->convertPhpToXml((float) 24.700);
+
+        $this->assertSame($xml, $output);
+    }
+}
```
veewee: Can you add a newline at the EOF?

veewee:
Hi @HKandulla,

Thanks for the PR! I've added some small remarks.
You can also add a double converter here.

src/Phpro/SoapClient/Soap/TypeConverter/DecimalTypeConverter.php
HKandulla: I was thinking about this too. I think null should be possible in all converter (no value) and strict types should always be nullable (?float), unfortunately this would make PHP 7.1 a requirement. In my case the service uses null to reset the value to its default. What do you think?

veewee: What do you mean with "reset the value to its default"?

HKandulla:
HI @veewee,
thanks for your remarks. I make the changes and also add a double converter.

src/Phpro/SoapClient/Soap/TypeConverter/DecimalTypeConverter.php
HKandulla: The SOAP-Service I am working with resets a property to its default value if I the corresponding tag is empty. If the tag has the value 0.0 it would set this value (and not the default)..

veewee: For me the value `null` is also OK, but we should take this in consideration if we want to support strict types at some point. Let's keep it this way for now :)

HKandulla: great. Thanks.

src/Phpro/SoapClient/ClientBuilder.php
```diff
@@ -88,6 +88,7 @@ public function __construct(ClientFactoryInterface $clientFactory, $wsdl, array
         // Add default converters:
         $this->addTypeConverter(new TypeConverter\DateTimeTypeConverter());
         $this->addTypeConverter(new TypeConverter\DateTypeConverter());
+        $this->addTypeConverter(new TypeConverter\DecimalTypeConverter());
```
veewee: I guess we should also add the double type converter to the defaults?

HKandulla: oh yes, sorry. It's added.

veewee:
Thanks for the work already. It looks like there is one thing left to do: updating the documentation.
https://github.com/phpro/soap-client/blob/master/docs/type-converter.md#convert-soap-types

HKandulla:
It's a pleasure. It's a great library! Is there anything else I can do?

veewee:
@HKandulla, Sure you can. There is the issue you created about ignoring null values (https://github.com/phpro/soap-client/issues/34)
Besides that, you can pick any open issue with the label [help wanted](https://github.com/phpro/soap-client/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22) you want or add something you are still missing to the package.

HKandulla:
I am still busy with the tests on #34. I want to add proper tests that make sure that even nodes with empty child nodes are removed by my middleware. But I definetely do this this week.. I was thinking about adding support for sequences as in:

`<s:complexType name="ArrayOfSyncCustomerRecord">
  <s:sequence>
<s:element minOccurs="0" maxOccurs="unbounded" name="SyncCustomerRecord" nillable="true" type="tns:SyncCustomerRecord"/>
</s:sequence>
</s:complexType>
`
But I need to investigate a little more. Any maybe the IteratorInterface is a good enough solution. If you have am idea of what is needed as far as support for sequences/arrays goes, please let me know.

veewee:
For sequences, I always use the Iterator. It works well and you can just iterate over the results as you probably always want to do. Let's continue the discussion in #34.

PR #40, opened 2017-04-07
Fixed authorization header for HTTP Basic Auth

geoffreytran:
HTTP Basic Auth middleware isn't using the correct header name. It should be Authorization.


veewee:
Wow, that is painful :) Thanks for noticing!

PR #44, opened 2017-05-12
allow to get original soap type 

insekticid:
allow to get original soap type and use it in custom CodeGenerator\Assembler

external Soap provider is using .NET ticks as xs:long so I need to know what was original type in my custom Assambler code generator

### DIFF
--- a/src/Phpro/SoapClient/CodeGenerator/Model/Property.php
+++ b/src/Phpro/SoapClient/CodeGenerator/Model/Property.php
@@ -22,6 +22,11 @@ class Property
     private $type;
 
     /**
+     * @var string
+     */
+    private $originalType;
+
+    /**
      * Property constructor.
      *
      * @param string $name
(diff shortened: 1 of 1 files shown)


insekticid:
any news here?

veewee:
Hi @insekticid,

Thanks for the PR!

Sorry for the late response. I do agree that we need this feature but I am not sure it is implemented the right way: I am thinking of passing the actual type object to the property and use the xsdName instead. But this should be done without breaking BC. I'll have to dive into the code again when I got more time.

veewee:
@janvernieuwe,

To clarify: It would be best if the `Property::getType()` method returned a `Type` object of which you could fetch the `Type::getName()` method with a possible shorthand: `Property::getTypeName()`.
The Type should contain an originalType or rawType property that can be used.

This is what would cause a BC break. Also: we'll need to check if this is useable this way inside the assemblers / rules.

veewee:
Closing this one since it's been inactive for a while.
Since the soap metadata layer is added, we might need to change code generation accordingly in a later version.

PR #46, opened 2017-06-14
JsonSerializable Assembler

veewee:
This new assembler makes it possible to JSON serialize all SOAP types. This could be used to log smaller versions of the request / responses etc.


eduarguz:
How could I use that in my logs? Do I need a new `LogPlugin` ?

veewee:
@eduarguz : It was added for custom logging implementations. In the internal implementation we use print_r, but that might not be what you want. In that case you can write your own logger based on the event system and use json_encode instead. (Some loggers like zendframework/zend-log try that by default)

eduarguz:
Thank you for your answer, was very useful.

Issue #50, opened 2017-06-30
Make it possible to configure the visibility in the PropertyAssembler


veewee:
Currently the PropertyAssembler generates protected properties by default.
Since we mostly don't extend these types, it is a better idea to change the default to private.

```php
    /**
     * @var string
     */
    protected $someFilter;
```

This could be done by adding the visibility as a constructor parameter to the PropertyAssembler and changing the defaults in the Configuration file. This way you can still choose to use protected or even public if you're a real daredevil.

veewee:
Fixed in #53

PR #53, opened 2017-06-30
Made property visibility configurable

janvernieuwe:
PR for issue https://github.com/phpro/soap-client/issues/50
Default visibility has been changed from protected to private


veewee:
Thanks for the PR! Changed version to 0.5 since this PR is changing the protected to private (BC break).

PR #55, opened 2017-07-28
Add validator plugin

veewee:
It is possible to use the [Symfony validator component](https://symfony.com/doc/current/components/validator.html) to validate your request objects before sending them to the server.
Since some servers return very cryptographic errors, the validation of request components could save you a lot of time during development.

The validator plugin is activated automatically when you attach a `ValidatorInterface` to the `ClientBuilder`. It will hook in to the Request event and will throw a `RequestException` when your request object doesn't contain valid data.

No more crappy error messages from the soap server!

### DIFF
--- a/docs/plugins/validator.md
+++ b/docs/plugins/validator.md
@@ -1,0 +1,12 @@
+# Validator plugin
+
+It is possible to use the [Symfony validator component](https://symfony.com/doc/current/components/validator.html)
+to validate your request objects before sending them to the server.
+Since some servers return very cryptographic errors, 
+the validation of request components could save you a lot of time during development.
+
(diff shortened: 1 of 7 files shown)


docs/plugins/validator.md
```diff
@@ -0,0 +1,12 @@
+# Validator plugin
+
+It is possible to use the [Symfony validator component](https://symfony.com/doc/current/components/validator.html)
+to validate your request objects before sending them to the server.
+Since some servers return very cryptographic errors, 
+the validation of request components could save you a lot of time during development.
+
+The validator plugin is activated automatically when you attach a `ValidatorInterface` to the `ClientBuilder`.
```
janvernieuwe: Maybe mention that this is done trough the withValidator method

PR #56, opened 2017-07-28
Client generator

janvernieuwe:
Ready for review


src/Phpro/SoapClient/CodeGenerator/Model/Client.php
```diff
@@ -0,0 +1,68 @@
...
+    /**
+     * TypeModel constructor.
+     *
+     * @param string $name
+     * @param string $namespace
+     * @param ClientMethodMap $methods
+     * @internal param string $xsdName
+     * @internal param Property[] $properties
+     */
+    public function __construct($name, $namespace, ClientMethodMap $methods)
```
veewee: You can use strict string types here

src/Phpro/SoapClient/CodeGenerator/Model/ClientMethod.php
```diff
@@ -0,0 +1,123 @@
...
+ * Class ClientMethod
+ *
+ * @package Phpro\SoapClient\CodeGenerator\Model
+ */
+class ClientMethod
+{
+    /**
+     * @var Method[]
+     */
+    private $parameters;
```
veewee: The type is Parameter[] instead of Method[]

src/Phpro/SoapClient/CodeGenerator/Model/ClientMethod.php
```diff
@@ -0,0 +1,123 @@
...
+
+    /**
+     * TypeModel constructor.
+     *
+     * @param $functionString
+     * @param $parameterNamespace
+     * @internal param string $xsdName
+     * @internal param Property[] $properties
+     */
+    public function __construct($functionString, $parameterNamespace = null)
```
veewee: strict types. Instead of null you can use an empty string

src/Phpro/SoapClient/CodeGenerator/Model/ClientMethod.php
```diff
@@ -0,0 +1,123 @@
...
+    {
+        $this->parameterNamespace = $parameterNamespace ?: '';
+        $this->parameters = $this->parseParameters($functionString);
```
veewee: The parsing should not occur in the constructor.
You can create a names constructor for this or maybe it's even better to move the parsing to a separate class or to the SoapClient class. This last one makes the package more consistent.

src/Phpro/SoapClient/CodeGenerator/Model/ClientMethod.php
```diff
@@ -0,0 +1,123 @@
...
+        $this->returnType = $this->parseReturnType($functionString);
+    }
+
+    /**
+     * @param $str
+     * @return array
+     */
+    private function parseParameters($str)
+    {
+        preg_match(
```
veewee: This could be on one line

src/Phpro/SoapClient/CodeGenerator/Model/ClientMethod.php
```diff
@@ -0,0 +1,123 @@
...
+            '/\((.*)\)/',
+            $str,
+            $properties
+        );
+
+        $parameters = [];
+        $properties = explode(',', $properties[1]);
```
veewee: You could do this explode inside the regex which is faster

src/Phpro/SoapClient/CodeGenerator/Model/ClientMethod.php
```diff
@@ -0,0 +1,123 @@
...
+        foreach ($properties as $property) {
+            list($type) = explode(' ', $property);
+            $parameters[] = new Parameter($type, $this->parameterNamespace.'\\'.$type);
```
veewee: What is the parameter namespace for? The model should be as close to soap as possible. The namespace shouldn't be here I think

src/Phpro/SoapClient/CodeGenerator/Model/ClientMethodMap.php
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * TypeMap constructor.
+     *
+     * @param array $methods
+     * @param string $parameterNamespace
+     */
+    public function __construct(array $methods, $parameterNamespace = null)
+    {
+        foreach ($methods as $method) {
```
veewee: It's probably better to move this logic to the names constructor instead?

src/Phpro/SoapClient/CodeGenerator/Model/ClientMethodMap.php
```diff
@@ -0,0 +1,50 @@
...
+            $this->methods[] = new ClientMethod($method, $parameterNamespace);
+        }
+    }
+
+    /**
+     * @param SoapClient $client
+     * @param string $parameterNamespace
+     * @return ClientMethodMap
+     */
+    public static function fromSoapClient(SoapClient $client, $parameterNamespace = null)
```
veewee: `string $parameterNamespace = '' `

src/Phpro/SoapClient/Console/Command/GenerateClientCommand.php
```diff
@@ -0,0 +1,220 @@
...
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ConfirmationQuestion;
+use Zend\Code\Generator\FileGenerator;
+
+/**
+ * Class GenerateClientCommand
+ *
+ * @package Phpro\SoapClient\Console\Command
+ */
+class GenerateClientCommand extends Command
```
veewee: This one looks hard on the type generator command. Maybe we can move some logic to helpers?
Maybe we can also rethink the commands a bit - since I do not think they are working 100% like they should.

src/Phpro/SoapClient/CodeGenerator/Assembler/ClientMethodAssembler.php
```diff
@@ -0,0 +1,60 @@
...
+                MethodGenerator::fromArray(
+                    [
+                        'name'       => $method->getMethodName(),
+                        'parameters' => $method->getParameters(),
+                        'visibility' => MethodGenerator::VISIBILITY_PUBLIC,
+                        'body'       => sprintf(
+                            'return $this->call(\'%1$s\', $%1$s);',
+                            $param->getName()
+                        ),
+                        'returntype'    => '\\'.$method->getParameterNamespace().'\\'.$method->getReturnType(),
```
veewee: This should be normalized. @tombevers  is working on a TypeChecker in #61

src/Phpro/SoapClient/CodeGenerator/Model/ClientMethod.php
janvernieuwe: This is the type namespace

src/Phpro/SoapClient/Console/Command/GenerateClientCommand.php
janvernieuwe: I already have so many changes, I'd prefer it if we can do this in a separate issue.

src/Phpro/SoapClient/CodeGenerator/Assembler/ClientMethodAssembler.php
janvernieuwe: Should i wait on Tom's PR?

PR #58, opened 2017-08-16
Use late static binding for the named constructors in XML class.

### LINKED ISSUE #57: `SoapXml::fromStream()` returns wrong object
The named constructor `SoapXml::fromStream()` does not return a `Phpro\SoapClient\Xml\SoapXml` object, but rather a `Phpro\SoapClient\Xml\Xml` object.

This makes it impossible to call any of the extended methods that the `SoapXml` class provides.

As an example, on the object you receive, you can't call a method like `getHeaders()`.

The problem is caused here:
https://github.com/phpro/soap-client/blob/02a7fb6faf306f9e1fc89c7b8651a13db2f352c4/src/Phpro/SoapClient/Xml/Xml.php#L93

This line should use late static binding to actually create an instance of the subclass that was requesting it, not the base `Xml` class:
```php
return new static($xml);
```

schlessera:
This makes sure the named constructors return the correct object when called from a subclass.

Fixes #57

### DIFF
--- a/src/Phpro/SoapClient/Xml/Xml.php
+++ b/src/Phpro/SoapClient/Xml/Xml.php
@@ -90,7 +90,7 @@ class Xml
         $xml = new DOMDocument();
         $xml->loadXML($stream->getContents());
 
-        return new self($xml);
+        return new static($xml);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


veewee:
Thanks for noticing and adding a PR!

PR #59, opened 2017-08-16
Adapt to immutable PSR-7 method `withBody()`

schlessera:
The example code for the middleware XML manipulation contains a wrong PSR-7 method.


veewee:
Thanks for the PR

Issue #49, opened 2017-06-30
Type hints for setter assemblers


janvernieuwe:
At the moment the setters have no type hints:
```
    /**
     * @param CardOrder $OrderInformation
     */
    public function setOrderInformation($OrderInformation)
    {
        $this->OrderInformation = $OrderInformation;
    }
```
It would be nice to enforce the types in the setters like so.
```
    /**
     * @param CardOrder $OrderInformation
     */
    public function setOrderInformation(CardOrder $OrderInformation)
    {
        $this->OrderInformation = $OrderInformation;
    }
```

Issue #36, opened 2017-02-17
Support return types


GaryJones:
A generated method like:

```php
    /**
     * @return string
     */
    public function getNewUsername()
    {
        return $this->NewUsername;
    }
```

could have a return type, if a config flag was set, so that it generated it as:

```php
   /**
     * @return string
     */
    public function getNewUsername(): string
    {
        return $this->NewUsername;
    }
```

veewee:
Small note: this might be a very tricke feature. Ext-soap does not support XSD simpletypes. If you create a named simpletype with type string, you'll see the name instead of the type string.

veewee:
Fixed in #61

PR #61, opened 2017-08-28
WIP: Adding return types

tombevers:
This PR will try to fix https://github.com/phpro/soap-client/issues/49 and https://github.com/phpro/soap-client/issues/36

Note this PR is still a WIP!


src/Phpro/SoapClient/CodeGenerator/Assembler/GetterAssemblerOptions.php
```diff
@@ -0,0 +1,73 @@
...
+    {
+        return new self();
+    }
+
+    /**
+     * @param bool $boolGetters
+     *
+     * @return GetterAssemblerOptions
+     */
+    public function withBoolGetters(bool $boolGetters): GetterAssemblerOptions
```
veewee: Since the property is false by default, you can set this one true by default

src/Phpro/SoapClient/CodeGenerator/Assembler/GetterAssemblerOptions.php
```diff
@@ -0,0 +1,73 @@
...
+
+        return $new;
+    }
+
+    /**
+     * @param bool $returnType
+     *
+     * @return GetterAssemblerOptions
+     */
+    public function withReturnType(bool $returnType): GetterAssemblerOptions
```
veewee: Since the property is false by default, you can set this one true by default

src/Phpro/SoapClient/CodeGenerator/Util/TypeChecker.php
```diff
@@ -0,0 +1,56 @@
...
+     * @link http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration
+     */
+    private static $internalPhpTypes = ['void', 'int', 'float', 'string', 'bool', 'array', 'callable', 'iterable'];
+
+    /**
+     * @param $type
+     *
+     * @return bool
+     */
+    public static function hasValidType(string $type): bool
```
veewee: Maybe the name could be something like `isKnownType()`.

src/Phpro/SoapClient/CodeGenerator/Util/TypeChecker.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+        if (self::isInternalPhpType($type) || self::isClassType($type)) {
```
veewee: You can just rerturn this instead of doing an if

src/Phpro/SoapClient/CodeGenerator/Assembler/FluentSetterAssembler.php
```diff
@@ -18,6 +20,21 @@
...
+     * @var FluentSetterAssemblerOptions
+     */
+    private $options;
+
+    /**
+     * FluentSetterAssembler constructor.
+     *
+     * @param FluentSetterAssemblerOptions $options
+     */
+    public function __construct(FluentSetterAssemblerOptions $options)
```
veewee: This will be a BC break. Should we make it possible to leave it empty?
I rather have it explicit, but since this is mostly configuration we might add an empty one manually when there are no options?

src/Phpro/SoapClient/CodeGenerator/Assembler/FluentSetterAssembler.php
```diff
@@ -60,4 +77,19 @@ public function assemble(ContextInterface $context)
...
+    /**
+     * @param Property $property
+     *
+     * @return array
+     */
+    private function getParameter(Property $property): array
+    {
+        $type = $property->getType();
+        if ($this->options->useReturnType() && TypeChecker::hasValidType($type)) {
+            return [['name' => $property->getName(), 'type' => $type]];
```
veewee: It's easier to debug when the keys start on a new line

src/Phpro/SoapClient/CodeGenerator/Assembler/FluentSetterAssemblerOptions.php
```diff
@@ -0,0 +1,47 @@
...
+    {
+        return new self();
+    }
+
+    /**
+     * @param bool $returnType
+     *
+     * @return FluentSetterAssemblerOptions
+     */
+    public function withReturnType(bool $returnType): FluentSetterAssemblerOptions
```
veewee: Since the property is false by default, you can set this one true by default

src/Phpro/SoapClient/CodeGenerator/Assembler/FluentSetterAssembler.php
janvernieuwe: This is for the parameters, not the method itself, I removed this part

src/Phpro/SoapClient/CodeGenerator/Util/TypeChecker.php
janvernieuwe: Or should we use the ones from zend-code and the function?
The one in zend-code is not complete now though (no self, iterable)

PR #62, opened 2017-09-01
Add test for WSSE encryption

### LINKED ISSUE #24: Add test for WSSE encryption
This is a follow-up ticket for #23:
WSSE encryption is not possible in newer versions of PHP: https://github.com/robrichards/wse-php/issues/28
This should be fixed once it is solved in the main package.

veewee:
This PR fixes #24


src/Phpro/SoapClient/Middleware/WsseMiddleware.php
```diff
@@ -156,6 +161,18 @@ public function withEncryption($serverCertificateFile)
     }
 
     /**
+     * @param bool $hasSubjectKeyIdentifier
+     *
+     * @return $this
+     */
+    public function withServerCertificateHasSubjectKeyIdentifier(bool $hasSubjectKeyIdentifier)
```
janvernieuwe: Shouldn't this be immutable?

veewee: It should. But the whole WsseMiddleware currently isn't written that way. I am going to leave it for now, but that's something we should surely handle at some point.

PR #64, opened 2017-09-04
Feature/make command more flexible

cjprinse:
make the commands GenerateClassmapCommand and GenerateTypesCommand extensible.

Main purpose of this feature is to setup a custom SoapClient (instead of just `new SoapClient()`), because we need a custom wsdl handler, since a webservice we are using does not have a valid certificate.

This is now possible by use the two new functions in the Config class: setGenerateTypesCommandClassName and setGenerateClassmapCommandClassName.
Here you can set your own command, which eventually will extends your base commands.

### DIFF
--- a/src/Phpro/SoapClient/CodeGenerator/Config/Config.php
+++ b/src/Phpro/SoapClient/CodeGenerator/Config/Config.php
@@ -107,14 +107,14 @@ class Config implements ConfigInterface
      */
     public function getWsdl()
     {
+        if (!$this->wsdl) {
+            throw InvalidArgumentException::wsdlConfigurationIsMissing();
+        }
+
         try {
             $wsdl = $this->wsdlProvider->provide($this->wsdl);
         } catch (WsdlException $e) {
             throw InvalidArgumentException::wsdlConfigurationIsMissing($e);
-        }
-
-        if (!$wsdl) {
-            throw InvalidArgumentException::wsdlConfigurationIsMissing();
         }
 
         return $wsdl;
--- a/src/Phpro/SoapClient/CodeGenerator/Config/Config.php
+++ b/src/Phpro/SoapClient/CodeGenerator/Config/Config.php
@@ -8,6 +8,8 @@ use Phpro\SoapClient\CodeGenerator\Rules\RuleInterface;
 use Phpro\SoapClient\CodeGenerator\Rules\RuleSet;
 use Phpro\SoapClient\CodeGenerator\Rules\RuleSetInterface;
 use Phpro\SoapClient\CodeGenerator\Util\Normalizer;
+use Phpro\SoapClient\Console\Command\GenerateClassmapCommand;
+use Phpro\SoapClient\Console\Command\GenerateTypesCommand;
 use Phpro\SoapClient\Exception\InvalidArgumentException;
 
 /**
@@ -47,6 +49,16 @@ class Config implements ConfigInterface
      * @var RuleSetInterface
      */
     protected $ruleSet;
+
+    /**
+     * @var string
+     */
+    protected $generateTypesCommandClassName = GenerateTypesCommand::class;
+
+    /**
+     * @var string
+     */
+    protected $GenerateClassmapCommandClassName = GenerateClassmapCommand::class;
 
     /**
      * Config constructor.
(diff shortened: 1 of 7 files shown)


veewee:
Hi @cjprinse,

Thanks for creating this PR.
Since this will make it possible for users to change the commands which could lead to very strange issues, I am not willing to support this feature.

Wouldn't it be a better idea to make the [WsdlProvider](https://github.com/phpro/soap-client/blob/master/src/Phpro/SoapClient/Wsdl/Provider/WsdlProviderInterface.php) and optionally the [SoapClientFactory](https://github.com/phpro/soap-client/blob/master/src/Phpro/SoapClient/Soap/SoapClientFactory.php) configurable (with sensible defaults) and use that one in the command?

cjprinse:
I can follow your thoughts. That's a fair point. Although I think it is also strange to add the SoapClientFactory in the config, along with the wsdl and soapoptions. Since these are already in the config.
Isn't it a better idea to extend the Config with the method `setClientBuilderConfig(ClientBuilderConfig $config)`? If set, we can make a ClientBuilder from the ClientBuilderConfig. I can refactor this PR is you think this is the correct direction.

veewee:
@cjprinse The problem with the ClientBuilder is that it creates a Client and not a SoapClient. It's probably best to make the WsdlProvider configurable with a default of MixedWsdlProvider. That way you can call the `WsdlProvider::provide()` method in $config->getWsdl(). That should be sufficient.

cjprinse:
@veewee 
now I'm getting the wsdl, but now the problem is that `$client->getSoapTypes() === null` in CodeGenerator/Model/TypeMap.php:49 is null. I'm guessing that since the $client does not get any configuration, so it won't accept invalid ssl by default

cjprinse:
@veewee 
It looks like the property $types in `src/Phpro/SoapClient/Soap/SoapClient.php` is null by default. Shouldn't it be an empty array?

veewee:
@cjprinse 
That should indeed be an array. Probably SoapClient::__getTypes() returns an empty array because of an invalid WSDL. 
Can you try downloading the WSDL manually and use that local file as wsdl? If that doesn't work, it's probably because it has to fetch additional resources which are behind the insecure connection.

You could try to use the SoapClient `stream_context` option with something like this to see if it works for you:
```
$context = stream_context_create([
    'ssl' => [
        // set some SSL/TLS specific options
        'verify_peer' => false,
        'verify_peer_name' => false,
        'allow_self_signed' => true
    ]
])
```

(This is not recommended on production by the way :) )

cjprinse:
@veewee 
Thanks, got it working!
PR seems to work ;)

src/Phpro/SoapClient/CodeGenerator/Config/Config.php
```diff
@@ -98,11 +107,17 @@ public function setNamespace($namespace)
...
     public function getWsdl()
     {
-        if (!$this->wsdl) {
+        try {
+            $wsdl = $this->wsdlProvider->provide($this->wsdl);
+        } catch (WsdlException $e) {
+            throw InvalidArgumentException::wsdlConfigurationIsMissing($e);
+        }
+
+        if (!$wsdl) {
```
veewee: Shouldn't we throw an error when the WSDL is empty before trying to load it through a WSDL provider?

src/Phpro/SoapClient/CodeGenerator/Config/Config.php
```diff
@@ -202,4 +217,23 @@ public function setRuleSet(RuleSetInterface $ruleSet)
         $this->ruleSet = $ruleSet;
         return $this;
     }
+
+    /**
+     * @return WsdlProviderInterface
+     */
+    public function getWsdlProvider()
```
veewee: Can you add this method to the ConfigInterface as well?
You might also want to add strict return type.

src/Phpro/SoapClient/CodeGenerator/Config/Config.php
```diff
@@ -202,4 +217,23 @@ public function setRuleSet(RuleSetInterface $ruleSet)
...
+    {
+        return $this->wsdlProvider;
+    }
+
+    /**
+     * @param WsdlProviderInterface $wsdlProvider
+     * @return Config
+     */
+    public function setWsdlProvider($wsdlProvider)
```
veewee: Can you add the parameter type to the method?

src/Phpro/SoapClient/CodeGenerator/Config/Config.php
cjprinse: not sure, maybe the returning wsdl can be empty? By example when using custom wsdl provider.

veewee: It should return a string or throw an exception. In any case: it doesnt really matter since it will throw an exception at some point. Lets leave it like this for now.

cjprinse:
@veewee bump

src/Phpro/SoapClient/CodeGenerator/Config/ConfigInterface.php
```diff
@@ -35,4 +36,9 @@ public function getDestination();
      * @return RuleSetInterface
      */
     public function getRuleSet();
+
+    /**
+     * @return WsdlProviderInterface
+     */
+    public function getWsdlProvider();
```
veewee: Missing `: WsdlProviderInterface`

veewee:
Hi @cjprinse,

Looks good. Can you also add tests for the changed code? Thanks

cjprinse:
@veewee  I've added some tests, but I don't know how to cover the other changes in phpspec

veewee:
Hi @cjprinse,

I'll add some tests by the end of the month. Sorry for the late answer :)

PR #67, opened 2017-09-15
Don't generate use statement for the same namespace

mheki:
Fixes `UseAssembler` preventing from generating `use` statement for the same namespace as of the current class.

### DIFF
--- a/src/Phpro/SoapClient/CodeGenerator/Assembler/UseAssembler.php
+++ b/src/Phpro/SoapClient/CodeGenerator/Assembler/UseAssembler.php
@@ -7,6 +7,7 @@ use Phpro\SoapClient\CodeGenerator\Context\PropertyContext;
 use Phpro\SoapClient\CodeGenerator\Context\TypeContext;
 use Phpro\SoapClient\CodeGenerator\Util\Normalizer;
 use Phpro\SoapClient\Exception\AssemblerException;
+use Zend\Code\Generator\ClassGenerator;
 
 /**
  * Class UseAssembler
@@ -51,6 +52,10 @@ class UseAssembler implements AssemblerInterface
     {
         $class = $context->getClass();
 
+        if ($this->usesTheSameNamespace($class)) {
+            return;
+        }
+
         try {
             $uses = $class->getUses();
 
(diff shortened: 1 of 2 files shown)


src/Phpro/SoapClient/CodeGenerator/Assembler/UseAssembler.php
```diff
@@ -63,4 +68,13 @@ public function assemble(ContextInterface $context)
...
         }
     }
+
+    /**
+     * @param ClassGenerator $class
+     * @return bool
+     */
+    private function usesTheSameNamespace(ClassGenerator $class)
+    {
+        return $this->useName === $class->getNamespaceName();
```
veewee: In a lot of cases, the `$this->useName` will be a fully qualified class name. 
E.g.:

```php
$assembler = new UseAssembler('MyNamespace\SomeOtherClass');
$class = new ClassGenerator('SomeClass', 'MyNamespace');
$context = new TypeContext($class, new Type(...));
```

In this case the result will be:

```php
namespace MyNamespace;

use MyNamespace\SomeOtherClass;

class SomeClass
{
}
```

Which leads to an unnecessary import since the SomeOtherClass is in the same namespace as SomeClass. PHP does not need a `use` statement for this. 

It's probably best to check on exact namespace or parent namespace?

test/PhproTest/SoapClient/Unit/CodeGenerator/Assembler/UseAssemblerTest.php
```diff
@@ -130,6 +130,30 @@ class MyType
     }
 
     /**
+     * @test
+     */
+    function it_does_not_assemble_use_for_the_same_namespace()
+    {
+        $assembler = new UseAssembler('MyNamespace');
```
veewee: As mentioned above: there should be 2 tests: one for exact namespace and one for equal parent namespace.

mheki: Thanks for the feedback, changes are here: https://github.com/phpro/soap-client/pull/67/commits/8c7cfd70e9622ac91acf79447eb3a641f9c44a6c
I've added few more tests with additional use cases.

veewee:
Thanks for this addition @mheki  !

PR #68, opened 2017-09-15
optimizes shell command to use soap-client

tknuppe:
The original entry is misleading. Noone knows where this soap-client resides.


veewee:
Thanks for the PR!

PR #71, opened 2017-09-22
Replace underscores with DIRECTORY_SEPARATOR (PSR-O, PSR-4)

janvernieuwe:
Replace underscores in classnames with / for PSR-0 and 4 when generating a filepath.


janvernieuwe:
Paths that do not exist are an issue

src/Phpro/SoapClient/Console/Command/GenerateTypesCommand.php
```diff
@@ -114,6 +114,8 @@ protected function execute(InputInterface $input, OutputInterface $output)
      */
     protected function handleType(TypeGenerator $generator, Type $type, $path)
     {
+        // Generate type sub folders if needed
+        $this->filesystem->ensureDirectoryExists(dirname($path));
```
veewee: We might want to transform $path to a SplFileInfo class so that we can avoid `dirname()`

test/PhproTest/SoapClient/Unit/CodeGenerator/Assembler/TypeModelTest.php
```diff
@@ -0,0 +1,26 @@
...
+namespace PhproTest\SoapClient\Unit\CodeGenerator\Assembler;
+
+use Phpro\SoapClient\CodeGenerator\Model\Type;
+
+/**
+ * Class UseAssemblerTest
+ *
+ * @package PhproTest\SoapClient\Unit\CodeGenerator\Assembler
+ */
+class TypeModelTest extends \PHPUnit_Framework_TestCase
```
veewee: It's better to remove this phpunit test and add the test inside the phpspec TypeSpec instead.

test/PhproTest/SoapClient/Unit/CodeGenerator/Assembler/TypeModelTest.php
```diff
@@ -0,0 +1,26 @@
...
+{
+
+    /**
+     * @test
+     */
+    function it_should_convert_underscores_in_path()
+    {
+        $assembler = new Type('\Foo\Bar', 'Bar_3_2', []);
```
veewee: It's not an assembler

spec/Phpro/SoapClient/CodeGenerator/Model/TypeSpec.php
```diff
@@ -17,7 +17,7 @@ class TypeSpec extends ObjectBehavior
 {
     function let()
     {
-        $this->beConstructedWith('MyNamespace', 'myType', ['prop1' => 'string']);
+        $this->beConstructedWith('MyNamespace', 'myType_3_2', ['prop1' => 'string']);
```
veewee: It's probably better to use a normal type name in the normal cases but pass a different name in the path_name case

src/Phpro/SoapClient/CodeGenerator/Model/Type.php
```diff
@@ -76,14 +76,25 @@ public function getXsdName()
     }
 
     /**
-     * @param $destination
-     *
+     * @param string $destination
+     * @return \SplFileInfo
+     */
+    public function getFileInfo($destination)
```
veewee: You can specify strict types here for both parameters as return type

src/Phpro/SoapClient/CodeGenerator/Model/Type.php
```diff
@@ -76,14 +76,25 @@ public function getXsdName()
...
+    {
+        $path = rtrim($destination, '/\\').DIRECTORY_SEPARATOR.$this->getName().'.php';
+        $path = str_replace('_', DIRECTORY_SEPARATOR, $path);
+
+        return new \SplFileInfo($path);
```
veewee: You can import this as use statement

src/Phpro/SoapClient/Util/Filesystem.php
```diff
@@ -14,10 +15,11 @@ class Filesystem
 {
 
     /**
-     * @param string $directory
+     * @param SplFileInfo $fileInfo
      */
-    public function ensureDirectoryExists($directory)
+    public function ensureDirectoryExists(SplFileInfo $fileInfo)
```
veewee: Since the rest of the class is using a string, we might as well keep it consistent :)

src/Phpro/SoapClient/Console/Command/GenerateTypesCommand.php
```diff
@@ -106,19 +109,18 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+    protected function handleType(TypeGenerator $generator, Type $type, SplFileInfo $fileInfo)
     {
         // Generate type sub folders if needed
-        $this->filesystem->ensureDirectoryExists(dirname($path));
+        $this->filesystem->ensureDirectoryExists($fileInfo);
         // Handle existing class:
-        if ($this->filesystem->fileExists($path)) {
-            if ($this->handleExistingFile($generator, $type, $path)) {
+        if ($fileInfo->isFile()) {
+            if ($this->handleExistingFile($generator, $type, $fileInfo->getPathname())) {
```
veewee: Would it be a lot of work to make the command work with the splFileInfo instead of converting it to a string here?

PR #74, opened 2017-09-29
Make sure to replace exact matching types only

### LINKED ISSUE #73: CodeGenerator - date changed to dateTime within a class name
In generated classes I can see:

```
    /**
     * @var ArrayOfConsoli\DateTimedAgreement
     */
    private $ConsolidatedFrom = null;
```

instead of:
```
    /**
     * @var ArrayOfConsolidatedAgreement
     */
    private $ConsolidatedFrom = null;
```

It looks like all doc block where `date` word is found are wrong with `\DateTime` replacing `date` in the original text.

veewee:
Fixes #73

### DIFF
--- a/src/Phpro/SoapClient/CodeGenerator/Util/Normalizer.php
+++ b/src/Phpro/SoapClient/CodeGenerator/Util/Normalizer.php
@@ -50,23 +50,16 @@ class Normalizer
         $normalizations = [
             'long' => 'int',
             'short' => 'int',
-            'dateTime' => '\\DateTime',
+            'datetime' => '\\DateTime',
             'date' => '\\DateTime',
             'boolean' => 'bool',
             'decimal' => 'float',
             'double' => 'float',
         ];
 
-        return preg_replace(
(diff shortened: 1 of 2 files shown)


src/Phpro/SoapClient/CodeGenerator/Util/Normalizer.php
```diff
@@ -47,15 +47,26 @@ public static function normalizeProperty($property)
...
+        return preg_replace(
+            array_map(
+                function ($search) {
+                    return '/^' . $search. '$/i';
+                },
+                array_keys($normalizations)
+            ),
+            array_values($normalizations),
+            $type
+        );
```
janvernieuwe: Perhaps this can be simplified a little bit by lowercasing the search key and using only lowercase in the array.
Check if the key exists and return it, otherwise return the input value.

veewee: Fixed

PR #77, opened 2017-09-29
Replace guzzle with httplug

### LINKED ISSUE #25: Use the http-interop layer instead of Guzzle.
This is a follow-up ticket for #23:

Currently we are using Guzzle as a hardcoded handler in this project.
This is because the http-interop doesn't have a standard on:

- What interfaces a http-client should have. (http-client)
- How a request / response should be created (http-factory)
- What client middlewares should look like (http-middleware)
- How async callbacks should work (async-interop/promise)

When all the groups above come to a deciscion, it is possible to make this package independent of the http-client you want to use. For now we'll only support the Guzzle implementations.

Actions:

- [ ] Keep track of the [PSR-15: http-middleware](https://github.com/php-fig/fig-standards/tree/master/proposed/http-middleware), [PSR-17 http-factory](https://github.com/php-fig/fig-standards/tree/master/proposed/http-factory) and [PSR-http-client](https://github.com/php-http/fig-standards/tree/master/proposed/http-client) changes to make sure that any php http-client can be used.
- [ ] Keep track of [guzzle changes](https://github.com/guzzle/guzzle/issues/1702) to the middlewares / http-factory and client PSRs.
- [ ] Keep track of the [async interop](https://github.com/async-interop/promise).

veewee:
fixes #25


veewee:
Upgrading:

```sh
composer require php-http/httplug:^1.0 php-http/message-factory:^1.0 php-http/discovery:^1.0 php-http/message:^1.0 php-http/guzzle6-adapter:^1.1
```

```php
// Wrap your guzzle client in a HTTPlug adapter:
$httpClient = new Client($guzzleClient);

// Use the new HttPlugHandle instead of the old GuzzleHandle
$clientBuilder->withHandler(HttPlugHandle::createForClient($httpClient));

// use the new HttPlugWsdlProvider instead of the old GuzzleWsdlHandle
$clientBuilder->withWsdlProvider(HttPlugWsdlProvider::createForClient($httpClient));

// If you are using NTLM:
// Remove the NtlmMiddleware and add the curl options to the guzzle client configuration instead:
$httpClient = Client::createWithConfig(['curl' => [
    CURLOPT_HTTPAUTH => CURLAUTH_NTLM,
    CURLOPT_USERPWD => 'username:password',
]]);
```

PR #83, opened 2017-10-09
Added a copyright disclaimer on all the src files

mosekwa:


mheki:
You can also add phpspec class templates for new classes. Example: https://github.com/phpspec/phpspec/blob/master/.phpspec/class.tpl

.idea/encodings.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
pierredup: The `.idea` folder should not be committed

composer.json
```diff
@@ -13,6 +13,7 @@
     "php": "^7.0",
     "doctrine/collections": "~1.3",
     "http-interop/http-factory": "^0.2.0",
+    "phpro/soap-client": "dev-master",
```
pierredup: This package should not depend on itself

bin/soap-client
```diff
@@ -1,6 +1,15 @@
 #!/usr/bin/env php
 <?php
 
+/*
+ * This file is part of the Phpro application.
```
pierredup: Should the name possibly be `Phpro SoapClient`, since phpro is just the vendor name?

.idea/encodings.xml
mosekwa: Noted, thanks. I will remove it and also add it to the git ignore file.

veewee:
We've dicussed this internally and decided not to add these copyright headers. Thanks for the PR anyways.

PR #85, opened 2017-10-13
Json serializable recursion

tknuppe:
This feature adds recusrion to the JsonSerializableAssembler. I don't know if it is complete but it works at least for me ;) Check it out and extend it if there is something missing or bad done. Thanks.


veewee:
FYI: When running json_encode, the method is being called recursively internally.

tknuppe:
Oh, ok. That's new to me.Thanks.

PR #91, opened 2017-10-19
Omit null

janvernieuwe:
~~DONT MERGE YET~~

To merge after zend-code has released a new version.
Update composer when this is done.


src/Phpro/SoapClient/CodeGenerator/Assembler/PropertyAssembler.php
```diff
@@ -51,21 +51,20 @@ public function assemble(ContextInterface $context)
...
+                    'tags' => [
+                        [
+                            'name'        => 'var',
+                            'description' => $property->getType(),
+                        ],
+                    ]
                 ])
-            );
+            ]);
+            $propertyGenerator->omitDefaultValue();
```
veewee: Isn't it possible to omit the default value with the `fromArray()` method?

janvernieuwe: Not yet, might add it.

janvernieuwe:
This can be merged, required php version needs to be upped to 7.1 though.

veewee:
Hi @janvernieuwe,

Can we talk about PHP versions later today? Give me a ping when you're ready for it ;)

janvernieuwe:
Just checked it, and the fromArray doesn't raise an error when the option is not defined.

test/PhproTest/SoapClient/Unit/CodeGenerator/Assembler/PropertyAssemblerTest.php
```diff
@@ -48,7 +48,8 @@ function it_assembles_a_property()
         $assembler->assemble($context);
 
         $code = $context->getClass()->generate();
-        $expected = <<<CODE
+        if (version_compare(PHP_VERSION, '7.1', '>=')) {
```
veewee: It's a better idea to create 2 separate tests and skip one of them based on the PHP version number.

PR #97, opened 2017-11-13
Auto namespace assembler

anuri73:
Usage:

`->addRule( new Rules\AssembleRule( new NamespaceAssembler( 'Klabs\Sale\AviaBundle\Service\Amadeus\Entity' ) ) )`

As parameter we can pass the prefix

### DIFF
--- a/src/Phpro/SoapClient/CodeGenerator/Assembler/NamespaceAssembler
+++ b/src/Phpro/SoapClient/CodeGenerator/Assembler/NamespaceAssembler
@@ -1,0 +1,81 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: Urmat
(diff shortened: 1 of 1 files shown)


veewee:
Hi @anuri73,

Thanks for the PR. Can you tell me why and how you want to use this assembler?
Currently the namespace of the models are automatically set by configuration.

References:
- https://github.com/phpro/soap-client/blob/master/src/Phpro/SoapClient/CodeGenerator/Config/ConfigInterface.php#L18
- https://github.com/phpro/soap-client/blob/master/src/Phpro/SoapClient/CodeGenerator/TypeGenerator.php#L43

veewee:
I assume this PR can be closed. Feel free to provide additional information about your case if you still need this generator.

PR #102, opened 2017-11-24
Fixes #36

### LINKED ISSUE #36: Support return types

janvernieuwe:
Return types were not yet enabled for getters in the types.
This PR fixes that.

There also was a a double method for the typeNamespace, this has been normalized to a single function that is in line with the others.

### DIFF
--- a/src/Phpro/SoapClient/CodeGenerator/Config/Config.php
+++ b/src/Phpro/SoapClient/CodeGenerator/Config/Config.php
@@ -80,7 +80,7 @@ class Config implements ConfigInterface
     public function __construct(string $wsdl = '', string $destination = '')
     {
         $this->setWsdl($wsdl);
-        $this->setDestination($destination);
+        $this->setTypeDestination($destination);
         $this->ruleSet = new RuleSet([
             new Rules\AssembleRule(new Assembler\PropertyAssembler()),
             new Rules\AssembleRule(new Assembler\ClassMapAssembler()),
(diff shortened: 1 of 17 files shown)


src/Phpro/SoapClient/CodeGenerator/Config/Config.php
```diff
@@ -115,17 +115,17 @@ public function getNamespace(): string
      */
     public function setNamespace(string $namespace): self
```
veewee: Feel free to remove these deprecated methods since we are breaking the contract anyways.

src/Phpro/SoapClient/CodeGenerator/Model/Property.php
```diff
@@ -21,16 +21,23 @@ class Property
...
+
     /**
      * Property constructor.
      *
      * @param string $name
      * @param string $type
+     * @param string $namespace
      */
-    public function __construct(string $name, string $type)
+    public function __construct(string $name, string $type, string $namespace = '')
```
veewee: Namespace should not be optional

src/Phpro/SoapClient/CodeGenerator/Config/Config.php
```diff
@@ -115,17 +115,17 @@ public function getNamespace(): string
...
+        $this->typeNamespace = Normalizer::normalizeNamespace($namespace);
 
         return $this;
     }
 
     /**
      * @return string
      */
-    public function getTypesNamespace()
+    public function getTypeNamespace()
```
veewee: You changed the method name but did not change the interface?

janvernieuwe: These were 2 methods one with and one without "s", getTypeNamespace is in the interface.

PR #107, opened 2017-12-08
Add option to override default converters - fix #106

### LINKED ISSUE #106: Can't override default type converters
Hi,
My problem is that I can not override the default type converters (datetime, date etc.) with a custom one because ClientBuilder::addTypeConverter() checking first that a converter already exists, and the default converters already added during ClientBuilder __constuct().

Solution: I think adding the default type converters on the build() method would solve this issue.

csiszarattila:
Hi,
this fixes #106.

To make this works first we move adding default converters at the end of building instead on construction but before we call factory on SoapClientFactory.
Therefore it makes possible to attach custom converters to replace built-in defaults.

### DIFF
--- a/src/Phpro/SoapClient/ClientBuilder.php
+++ b/src/Phpro/SoapClient/ClientBuilder.php
@@ -98,12 +98,6 @@ class ClientBuilder
         $this->clientFactory = $clientFactory;
         $this->wsdl = $wsdl;
         $this->soapOptions = $soapOptions;
-
-        // Add default converters:
-        $this->addTypeConverter(new TypeConverter\DateTimeTypeConverter());
-        $this->addTypeConverter(new TypeConverter\DateTypeConverter());
-        $this->addTypeConverter(new TypeConverter\DecimalTypeConverter());
-        $this->addTypeConverter(new TypeConverter\DoubleTypeConverter());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Phpro/SoapClient/ClientBuilder.php
```diff
@@ -170,6 +164,25 @@ public function addTypeConverter(TypeConverterInterface $typeConverter)
         $this->converters->add($typeConverter);
     }
 
+    /**
+     * Add the default to converters
+     */
+    public function addDefaultTypeConverters()
```
veewee: I think it's better to move this logic to the `TypeConverterCollection` class

src/Phpro/SoapClient/ClientBuilder.php
```diff
@@ -186,14 +199,26 @@ public function addMiddleware(MiddlewareInterface $middleware)
         $this->middlewares[] = $middleware;
     }
 
+    /**
+     * @return SoapClientFactory
+     */
+    public function createSoapClientFactory()
+    {
+        // Add default converters:
+        $this->addDefaultTypeConverters();
```
veewee: The defaults could be added in the factory instead of in the builder:
`\Phpro\SoapClient\Soap\SoapClientFactory::factory()`

This way we can make the defaults optional in a future version by setting a soapOption

veewee:
This will be fixed in #197.

PR #110, opened 2017-12-19
The most basic config generator

janvernieuwe:
Works for the very most basic configuration.
Do we want to be able to set the rules interactively?

Still needs documentation.


veewee:
Hi @janvernieuwe,

Maybe it's better to use the same structure as the other commands?
Input -> Context -> Generator -> file

This way we can reuse the file exists logic from the other commands once we extract that from the commands. OK: it's a bit overkill for what it does, but it makes the whole package easier to understand. What do you think?

Maybe you could also optionally ask for a response / request regex to add some basic rules to make the client work. Something like:
```php
->addRule(
        new Rules\TypenameMatchesRule(
            new Rules\MultiRule([
                new Rules\AssembleRule(new Assembler\ConstructorAssembler()),
            ]),
            '/Request$/i'
        )
    )
    // Add response rules:
    ->addRule(
        new Rules\TypenameMatchesRule(
            new Rules\MultiRule([
                new Rules\AssembleRule(new Assembler\ResponseAssembler()),
                new Rules\AssembleRule(new Assembler\GetterAssembler()),
            ]),
            '/Response$/i'
        )
    );
```

janvernieuwe:
Sure, I'll add a context and a generator.
I'll also look into adding the request and response rule (regex).

I'm not going to dynamically add the rules (read they are a fixed set, they can be edited manually once generated).

spec/Phpro/SoapClient/CodeGenerator/ConfigGeneratorSpec.php
```diff
@@ -0,0 +1,88 @@
...
+use Phpro\SoapClient\CodeGenerator\GeneratorInterface;
+use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
+use Phpro\SoapClient\CodeGenerator\ConfigGenerator;
+use Zend\Code\Generator\FileGenerator;
+
+/**
+ * Class ConfigGeneratorSpec
+ */
+class ConfigGeneratorSpec extends ObjectBehavior
```
veewee: Please keep the code generation tests in the tests directory so that they are all in one place

spec/Phpro/SoapClient/CodeGenerator/ConfigGeneratorSpec.php
```diff
@@ -0,0 +1,88 @@
...
+                'setTypeNamespace'       => 'App\\Type',
+                'setClientDestination'   => 'src/client',
+                'setClientName'          => 'Client',
+                'setClientNamespace'     => 'App\\Client',
+                'setClassmapDestination' => 'src/classmap',
+                'setClassmapName'        => 'Classmap',
+                'setClassmapNamespace'   => 'App\\Classmap',
+            ]
+        );
+        $context->getRequestRegex()->willReturn('/Request$/i');
```
veewee: Also provide test cases without request / response regex

src/Phpro/SoapClient/CodeGenerator/Context/ConfigContext.php
```diff
@@ -0,0 +1,65 @@
+<?php
+
+namespace Phpro\SoapClient\CodeGenerator\Context;
+
+class ConfigContext implements ContextInterface
+{
+    private $setters = [];
+    private $requestRegex = '/Request$/i';
```
veewee: It's probably not a good idea to add these defaults since it is depending on the server implementation.
Or how do you see this?

src/Phpro/SoapClient/CodeGenerator/Context/ConfigContext.php
```diff
@@ -0,0 +1,65 @@
...
+    private $responseRegex = '/Response$/i';
+
+    public function addSetter(string $name, string $value, bool $namespace = false): self
+    {
+        if ($value === '') {
```
veewee: Not sure why this check is added. This method shouldn't be called if the value is empty right?

src/Phpro/SoapClient/CodeGenerator/Context/ConfigContext.php
```diff
@@ -0,0 +1,65 @@
...
+            return $this;
+        }
+        if ($namespace) {
```
veewee: I understand why it is added here, but I would like to avoid this kind of logic inside this method

src/Phpro/SoapClient/CodeGenerator/Context/ConfigContext.php
```diff
@@ -0,0 +1,65 @@
...
+            $this->setters[$name] = str_replace('/', '\\\\', $value);
+
+            return $this;
+        }
+        $this->setters[$name] = addslashes($value);
```
veewee: Shouldn't the slashes be added while generating the code instead of altering the users input at this point?

src/Phpro/SoapClient/CodeGenerator/ConfigGenerator.php
```diff
@@ -0,0 +1,74 @@
...
+RULESET;
+
+
+    /**
+     * @param      $name
+     * @param      $value
+     */
+    private function addSetter($name, $value)
+    {
+        $this->body .= sprintf("\t->%s('%s')".PHP_EOL, $name, $value);
```
veewee: Please use `\Zend\Code\Generator\AbstractGenerator::$indentation` instead of the hardcoded tab here

src/Phpro/SoapClient/CodeGenerator/ConfigGenerator.php
```diff
@@ -0,0 +1,74 @@
...
+                    new Rules\AssembleRule(new Assembler\ConstructorAssembler()),
+                ]),
+                '%s'
+            )
+        )
+        ->addRule(
+            new Rules\TypenameMatchesRule(
+                new Rules\MultiRule([
+                    new Rules\AssembleRule(new Assembler\ResultAssembler()),
+                    new Rules\AssembleRule(new Assembler\GetterAssembler(\$getterOptions)),
```
veewee: It looks like a better idea to drop the variable here and just add the `new Assembler\GetterAssemblerOptions()` instead. It makes it easier to start configuring for new users.

src/Phpro/SoapClient/Console/Command/GenerateConfigCommand.php
```diff
@@ -0,0 +1,122 @@
...
+            ->setName(self::COMMAND_NAME)
+            ->setDescription('Interactively generate basic configuration');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $context = new ConfigContext();
+        $io = new SymfonyStyle($input, $output);
+        $io->section('Config settings');
+        $dest = $io->ask('config location (Where to put the config, including .php)', 'config/default.php');
```
veewee: Maybe use a better default config name like `config/soap-client.php` which tells the user what is inside the file?

src/Phpro/SoapClient/Console/Command/GenerateConfigCommand.php
```diff
@@ -0,0 +1,122 @@
...
+
+    protected function typeConfig(SymfonyStyle $io, ConfigContext $context)
+    {
+        $io->section('Type Configuration');
+        if (!$io->confirm('Do you want to configure the types?')) {
+            return;
+        }
+        $context->addSetter(
+            'setTypeDestination',
+            $io->ask('destination (location where files are generated)', 'src/type')
```
veewee: This default doesn't look autoloadable

src/Phpro/SoapClient/Console/Command/GenerateConfigCommand.php
```diff
@@ -0,0 +1,122 @@
...
+
+        $generator = new ConfigGenerator();
+        $this->filesystem->putFileContents($dest, $generator->generate(new FileGenerator(), $context));
+        $io->success("Config has been written to $dest");
+    }
+
+    protected function typeConfig(SymfonyStyle $io, ConfigContext $context)
+    {
+        $io->section('Type Configuration');
+        if (!$io->confirm('Do you want to configure the types?')) {
```
veewee: It won't be possible to generate types when this is skipped right? Maybe we should avoid to skip this or add the defaults at some other location?

src/Phpro/SoapClient/Console/Command/GenerateConfigCommand.php
```diff
@@ -0,0 +1,122 @@
...
+            'setTypeNamespace',
+            $io->ask('namespace (namespace of all types)', 'App/Type'),
+            true
+        );
+    }
+
+    protected function clientConfig(SymfonyStyle $io, ConfigContext $context)
+    {
+        $io->section('Client Configuration');
+        if (!$io->confirm('Do you want to configure the client?')) {
```
veewee: It won't be possible to generate the client when this is skipped right? Maybe we should avoid to skip this or add the defaults at some other location?

src/Phpro/SoapClient/Console/Command/GenerateConfigCommand.php
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->setName(self::COMMAND_NAME)
+            ->setDescription('Interactively generate basic configuration');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
```
veewee: Maybe it's easier to give the user less options.
For new users it might be easier just to provide a base destination folder and namespace and place the files in the directories we suggest.
If you want more advanced configuration options, you can always change the config file right?

src/Phpro/SoapClient/Console/Command/GenerateConfigCommand.php
```diff
@@ -0,0 +1,122 @@
...
+            'setClassmapNamespace',
+            $io->ask('namespace (namespace of the classmap)', 'App/Classmap'),
+            true
+        );
+    }
+
+    public function rulesetConfig(SymfonyStyle $io, ConfigContext $context)
+    {
+        $io->section('Ruleset Configuration');
+        $context->setRequestRegex($io->ask('Regex for matching request objects', $context->getRequestRegex()));
```
veewee: Asking for a regex might be hard for some users. Maybe we should ask for a keyword and generate a regex for that?

src/Phpro/SoapClient/CodeGenerator/Context/ConfigContext.php
janvernieuwe: I removed the defaults, if its empty the ruleset will be skipped on generation.

src/Phpro/SoapClient/Console/Command/GenerateConfigCommand.php
janvernieuwe: What directories do you suggest, I was thinking about:
BASE_DIR /Client
BASE_DIR /Type
BASE_DIR /Classmap

janvernieuwe: The classmap could perhaps also be put in the Client dir?

spec/Phpro/SoapClient/CodeGenerator/ConfigGeneratorSpec.php
janvernieuwe: Done

src/Phpro/SoapClient/Console/Command/GenerateConfigCommand.php
janvernieuwe: Copy pasta from chat:

BASEDIR = some/project/namespace/SomeClientName/

en dan:
- Type/*.php
- ClassMapFactory.php
- Client.php
- ClientFactory.php

janvernieuwe: This has been implemented. Please review.

docs/code-generation/configuration.md
```diff
@@ -30,6 +30,9 @@ return Config::create()
 ;
 ```
 
+Luckily a command is provided to generate this for you in an interactive manner.
```
veewee: Can you add some additional information on how to use the command exactly and what options can be used?
Example:
https://github.com/phpro/soap-client/blob/master/docs/cli/generate-types.md

src/Phpro/SoapClient/CodeGenerator/ConfigGenerator.php
```diff
@@ -0,0 +1,85 @@
...
+use Zend\Code\Generator\FileGenerator;
+
+/**
+ * Class ConfigGenerator
+ *
+ * @package Phpro\SoapClient\CodeGenerator
+ */
+class ConfigGenerator implements GeneratorInterface
+{
+    private $body = <<<BODY
```
veewee: These templates should be constants

src/Phpro/SoapClient/CodeGenerator/ConfigGenerator.php
```diff
@@ -0,0 +1,85 @@
...
+
+
+    /**
+     * @param               $name
+     * @param               $value
+     * @param FileGenerator $file
+     */
+    private function addSetter($name, $value, FileGenerator $file)
+    {
+        $this->body .= sprintf("%s->%s('%s')".PHP_EOL, $file->getIndentation(), $name, $value);
```
veewee: Instead of changing the generator's state, I would prefer to return the body

src/Phpro/SoapClient/CodeGenerator/ConfigGenerator.php
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /**
+     * @param FileGenerator $file
+     * @param ConfigContext $context
+     * @return string
+     */
+    public function generate(FileGenerator $file, $context): string
+    {
+        $file->setUse('Phpro\\SoapClient\\CodeGenerator\\Assembler');
```
veewee: This method should build the body in-memory instead of changing the generator's state. (Just like the other generators). This way a generator can be reused

src/Phpro/SoapClient/CodeGenerator/Context/ConfigContext.php
```diff
@@ -0,0 +1,71 @@
+<?php
+
+namespace Phpro\SoapClient\CodeGenerator\Context;
+
+class ConfigContext implements ContextInterface
```
veewee: Can you remove the set* methods and use the constructor instead?

src/Phpro/SoapClient/Console/Command/GenerateConfigCommand.php
```diff
@@ -0,0 +1,88 @@
...
+        $this->addNonEmptySetter($context, 'setClientName', $name.'Client');
+        $this->addNonEmptySetter($context, 'setClientNamespace', $namespace);
+
+        // Classmap
+        $this->addNonEmptySetter($context, 'setClassMapDestination', $baseDir);
+        $this->addNonEmptySetter($context, 'setClassMapName', $name.'Classmap');
+        $this->addNonEmptySetter($context, 'setClassMapNamespace', $namespace);
+
+        // Ruleset
+        $requestKeyword = $io->ask('Keyword for matching request objects', 'Request');
```
veewee: The problem here is that not all response or request objects contain a keyword. Any suggestions on how to fix this? Maybe you could ask if both request and response can be identifier by a keyword?

I've seen lots of implementations that look like:

- SomethingRequest ( = Request oject)
- Something ( = Response object)

docs/code-generation/configuration.md
janvernieuwe: Behind that text the command is given.
Since it's an interactive command, it has no real options.

I don't know what more i can tell about this in the documentation.

src/Phpro/SoapClient/Console/Command/GenerateConfigCommand.php
janvernieuwe: Hmyeah, we could ask for the whole regex like we did in at the start of this.

src/Phpro/SoapClient/CodeGenerator/Context/ConfigContext.php
janvernieuwe: This messes up the command, and the keywords are not required by the generator (it just skips the rule generation in that case)

src/Phpro/SoapClient/Console/Command/GenerateConfigCommand.php
veewee: As discussed : lets put the defaults to an empty string. 
This way you can choose: Add a keyword or no rule generation

PR #119, opened 2017-12-30
Update sample-config.php for 0.6

GaryJones:
Adds required GetterAssemblerOptions and avoids deprecated `setName()` and `setType()` methods.


veewee:
I did a quick search but couldn't find where this file is being used. Can you tell me if you found a link to this file or just bumped against it? I guess it can be removed completely instead.

GaryJones:
I couldn't find a link to it either - but didn't know if it was still used as a standalone reference.

veewee:
Fixed in https://github.com/phpro/soap-client/commit/486c3cb2e62d80e8d57299ec9d01eb0351650651

Issue #121, opened 2017-12-30
Generated Client method return types


GaryJones:
Here's the relevant bit of the WSDL:

```xml
<s:element name="GetReleaseNotes">
    <s:complexType>
        <s:sequence>
            <s:element minOccurs="0" maxOccurs="1" name="x" type="s:string" />
            <s:element minOccurs="0" maxOccurs="1" name="y" type="s:string" />
        </s:sequence>
    </s:complexType>
</s:element>
<s:element name="GetReleaseNotesResponse">
    <s:complexType>
        <s:sequence>
            <s:element minOccurs="0" maxOccurs="1" name="GetReleaseNotesResult" type="tns:GetReleaseNotesResult" />
        </s:sequence>
    </s:complexType>
</s:element>
<s:complexType name="GetReleaseNotesResult">
    <s:complexContent mixed="false">
        <s:extension base="tns:CorpWebResult">
            <s:sequence>
                <s:element minOccurs="0" maxOccurs="1" name="MajorVersions" type="tns:ArrayOfReleaseMajorVersion" />
            </s:sequence>
        </s:extension>
    </s:complexContent>
</s:complexType>
```

My understanding is, `GetReleaseNotesResponse` is a ResultProvider, and as such, I have the following Rule:

```php
->addRule(new TypenameMatchesRule(
    new AssembleRule(new ResultProviderAssembler()),
        '/Response$/'
))
```

This gives me a `GetReleaseNotesResponse` of:

```php
namespace GAA\CorpManagerApi\SoapTypes;

use Phpro\SoapClient\Type\ResultProviderInterface;
use Phpro\SoapClient\Type\ResultInterface;

class GetReleaseNotesResponse implements ResultProviderInterface
{

    /**
     * @var GetReleaseNotesResult
     */
    private $GetReleaseNotesResult = null;

    /**
     * @return GetReleaseNotesResult|ResultInterface
     */
    public function getResult() : \Phpro\SoapClient\Type\ResultInterface
    {
        return $this->GetReleaseNotesResult;
    }
}
```

Now, in the `\Phpro\SoapClient\Client::call()` method, there is [code](https://github.com/phpro/soap-client/blob/5e0f5dd74a0a261db8c73cb669bdc376ad258d50/src/Phpro/SoapClient/Client.php#L97-L99) to skip straight to the `getResult()` value:

```php
if ($result instanceof ResultProviderInterface) {
    $result = $result->getResult();
}
```

[...]

Issue #117, opened 2017-12-29
Add DocBlocks to output of generate:client


GaryJones:
The current output of the `generate:client` code has no documentation. Other than making it harder to follow, IDEs also complain:

![screenshot 2017-12-29 17 04 41](https://user-images.githubusercontent.com/88371/34442482-620dfbe0-ecba-11e7-82bd-9b08bc50ab5d.png)

When DocBlocks are added, with the right types for `@return`, then IDEs are happier (not only for this file, but for code that uses this Client):

![screenshot 2017-12-29 17 05 23](https://user-images.githubusercontent.com/88371/34442525-9d84ec42-ecba-11e7-9c34-7dcb3a06c975.png)

veewee:
Adding the docblocks shouldn't be hard.

We currently don't have access to the SOAP function description since we are using the built-in [SoapClient::__getFunctions()](http://php.net/manual/en/soapclient.getfunctions.php) method to load the function information.  As you can see in the docs, there is no description being returned there. However, this might be possible once #65 is ready.

GaryJones:
A starting point (though I'm sure it could be improved) for adding docs via `ClientMethodAssembler::assemble` and the `MethodGenerator::fromArray()` call:

```php
'docblock' => DocBlockGenerator::fromArray([
    'tags' => [
        [
            'name' => 'param',
            'description' => sprintf('\%s $%s', RequestInterface::class, 'request'),
        ],
        [
        	'name' => 'return',
            'description' => '\\' . ResultInterface::class . '|\\'.$method->getParameterNamespace().'\\'.$method->getReturnType(),
        ],
        [
        	'name' => 'throws',
            'description' => '\\' . SoapException::class,
        ],
    ]
]),
```

For me, this produces a method like:

![screenshot 2017-12-29 18 58 39](https://user-images.githubusercontent.com/88371/34445229-4ec84df0-ecca-11e7-96ea-bb0ab577eaec.png)

Which would be fine, except that the `@return` description is wrapped on the next line, and so isn't picked up.

Cutting the `return` description down to:

```php
'description' => '\\' . ResultInterface::class,
```
And we get a working version (happy IDE) of:

![screenshot 2017-12-29 19 01 21](https://user-images.githubusercontent.com/88371/34445287-ac169746-ecca-11e7-8b53-a3e00c53b144.png)

veewee:
Feel free to create a PR in which we can discuss the code.

PR #122, opened 2017-12-30
Client Generation: Add DocBlocks to methods

GaryJones:
First pass. 

Note that the documented return type, and the return type declarations are not currently the same, as it's possible that the return type may be wrong (#121).

WordWrap is disabled, otherwise the FQN of the classes results in tags that are invalid (or at least not recognised by PhpStorm):

```php
/**
 * @param
 * \Phpro\SoapClient\Type\RequestInterface|\GAA\CorpManagerApi\SoapTypes\IsBlocked $IsBlocked
```

Results in client methods like:

screenshot 2017-12-30 14 53 06

See #117

### DIFF
--- a/src/Phpro/SoapClient/CodeGenerator/Assembler/ClientMethodAssembler.php
+++ b/src/Phpro/SoapClient/CodeGenerator/Assembler/ClientMethodAssembler.php
@@ -7,6 +7,10 @@ use Phpro\SoapClient\CodeGenerator\Context\ContextInterface;
 use Phpro\SoapClient\CodeGenerator\Context\ClientMethodContext;
 use Phpro\SoapClient\CodeGenerator\Model\Parameter;
 use Phpro\SoapClient\Exception\AssemblerException;
+use Phpro\SoapClient\Exception\SoapException;
+use Phpro\SoapClient\Type\RequestInterface;
+use Phpro\SoapClient\Type\ResultInterface;
+use Zend\Code\Generator\DocBlockGenerator;
 use Zend\Code\Generator\MethodGenerator;
 
 /**
@@ -52,6 +56,27 @@ class ClientMethodAssembler implements AssemblerInterface
                         ),
                         // TODO: Use normalizer once https://github.com/phpro/soap-client/pull/61 is merged
                         'returntype' => '\\'.$method->getParameterNamespace().'\\'.$method->getReturnType(),
+                        'docblock' => DocBlockGenerator::fromArray([
+                            'tags' => [
+                                [
+                                    'name' => 'param',
+                                    'description' => sprintf(
+                                        '\%s|\%s $%s',
+                                        RequestInterface::class,
+                                        $param->getType(),
+                                        $param->getName()
+                                    ),
+                                ],
+                                [
+                                    'name' => 'return',
+                                    'description' => '\\' . ResultInterface::class,
+                                ],
+                                [
+                                    'name' => 'throws',
+                                    'description' => '\\' . SoapException::class,
+                                ],
+                            ]
+                        ])->setWordWrap(false),
                     ]
                 )
             );
--- a/src/Phpro/SoapClient/CodeGenerator/Assembler/ClientMethodAssembler.php
+++ b/src/Phpro/SoapClient/CodeGenerator/Assembler/ClientMethodAssembler.php
@@ -64,7 +64,7 @@ class ClientMethodAssembler implements AssemblerInterface
                                     'description' => sprintf(
                                         '%s|%s $%s',
                                         $this->generateShortClassNameAndAddImport(RequestInterface::class, $class),
-                                        $this->generateShortClassNameAndAddImport($param->getType(), $class),
+                                        $this->getPrefixedClassNameAndAddImport($param->getType(), $class),
                                         $param->getName()
                                     ),
                                 ],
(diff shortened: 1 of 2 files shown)


src/Phpro/SoapClient/CodeGenerator/Assembler/ClientMethodAssembler.php
```diff
@@ -52,6 +56,27 @@ public function assemble(ContextInterface $context): bool
                         ),
                         // TODO: Use normalizer once https://github.com/phpro/soap-client/pull/61 is merged
                         'returntype' => '\\'.$method->getParameterNamespace().'\\'.$method->getReturnType(),
+                        'docblock' => DocBlockGenerator::fromArray([
+                            'tags' => [
+                                [
+                                    'name' => 'param',
+                                    'description' => sprintf(
+                                        '\%s|\%s $%s',
```
veewee: You can also add the short classname instead of the full classname here and add it to the uses of the class. That way the docblocks contain the short name.
You could prefix the type so that you don't need to import all classes.

src/Phpro/SoapClient/CodeGenerator/Assembler/ClientMethodAssembler.php
```diff
@@ -52,6 +56,27 @@ public function assemble(ContextInterface $context): bool
...
+                                        RequestInterface::class,
+                                        $param->getType(),
+                                        $param->getName()
+                                    ),
+                                ],
+                                [
+                                    'name' => 'return',
+                                    'description' => '\\' . ResultInterface::class,
```
veewee: I prefer returning the actual return type instead of the interface in the implementation.

src/Phpro/SoapClient/CodeGenerator/Assembler/ClientMethodAssembler.php
```diff
@@ -52,6 +56,27 @@ public function assemble(ContextInterface $context): bool
...
+                                ],
+                                [
+                                    'name' => 'throws',
+                                    'description' => '\\' . SoapException::class,
```
veewee: Also here it seems better to use the short class name instead.

veewee:
Indeed, the return type can be invalid while using ResultProviders. We'll need a workaround for that in a different PR.

src/Phpro/SoapClient/CodeGenerator/Assembler/ClientMethodAssembler.php
GaryJones: > add it to the uses of the class

Any pointers on how to go about adding to the `use` statements please?

GaryJones: [`addUse()`](https://github.com/zendframework/zend-code/blob/6b1059db5b368db769e4392c6cb6cc139e56640d/src/Generator/ClassGenerator.php#L738).

src/Phpro/SoapClient/CodeGenerator/Assembler/ClientMethodAssembler.php
GaryJones: Until #121 is fixed, the `ResultInterface` is needed to keep IDEs happy.

screenshot 2018-01-04 11 45 33

Note the addition of the `ResultInterface` on the second method.

GaryJones:
The [latest commit](https://github.com/phpro/soap-client/pull/122/commits/f1a4998b51d988224bcc947da4239eb0e9230989) produces a nicer output:

screenshot 2018-01-04 12 15 55

I think the code organisation could be improved further though, but wanted to offer them here for discussion before opening new issues / updating the PR with out-of-scope changes. Thoughts:

- Update `Normalizer::getClassNameFromFQN` with an optional second argument (int), so that it allows the number of namespace prefix levels to leave attached to the class name. i.e.

```php
echo Normalizer::getClassNameFromFQN('My\Project\FooClass'); // "FooClass"
echo Normalizer::getClassNameFromFQN('My\Project\FooClass', 1); // "Project\FooClass"
echo Normalizer::getClassNameFromFQN('My\Project\FooClass', 99); // "My\Project\FooClass"
```
- Add in a new `Normalizer::getNamespaceFromFQN()` method which accepts a FQN string, and an integer level of how many parts to drop from the end. i.e.

```php
echo Normalizer::getNamespaceFromFQN('My\Project\FooClass'); // "My\Project\FooClass"
echo Normalizer::getClassNameFromFQN('My\Project\FooClass', 1); // "My\Project"
echo Normalizer::getClassNameFromFQN('My\Project\FooClass', 99); // ""
```

Both of these would make it easier to then generate short class names with prefixes, and would cut the number of potential imports down substantially:

screenshot 2018-01-04 12 25 43

GaryJones:
Non-related Travis failure.

GaryJones:
Another non-related Travis failure: https://travis-ci.org/phpro/soap-client/jobs/325019033

veewee:
Hi @GaryJones,

I wouldn't add a prefix for the RequestInterface and ResponseInterface types.
Next it is possible to add a default alias to an import. In this case, you know that e.g. `Type` is not taken, so you can always use this hardcoded prefix.

The suggestions for adding an int to the methods doesn't sound like a good idea at first sight since you mostly do not know how many namespace parts there are in advance so you'll have to calculate it. Therefor it seems better to create a method that does whatever you want to do instead.

GaryJones:
WET code, but added a `generatePrefixedClassNameAndAddImport()` method. Result is:

screenshot 2018-01-04 15 39 46

How does that output look?

PR #123, opened 2017-12-31
Fix return type for generated client methods

### LINKED ISSUE #121: Generated Client method return types

GaryJones:
Fixes #121

### DIFF
--- a/src/Phpro/SoapClient/CodeGenerator/Assembler/ClientMethodAssembler.php
+++ b/src/Phpro/SoapClient/CodeGenerator/Assembler/ClientMethodAssembler.php
@@ -7,6 +7,7 @@ use Phpro\SoapClient\CodeGenerator\Context\ContextInterface;
 use Phpro\SoapClient\CodeGenerator\Context\ClientMethodContext;
 use Phpro\SoapClient\CodeGenerator\Model\Parameter;
 use Phpro\SoapClient\Exception\AssemblerException;
+use Phpro\SoapClient\Type\ResultInterface;
 use Zend\Code\Generator\MethodGenerator;
 
 /**
@@ -50,8 +51,7 @@ class ClientMethodAssembler implements AssemblerInterface
                             'return $this->call(\'%1$s\', $%1$s);',
                             $param->getName()
                         ),
-                        // TODO: Use normalizer once https://github.com/phpro/soap-client/pull/61 is merged
-                        'returntype' => '\\'.$method->getParameterNamespace().'\\'.$method->getReturnType(),
+                        'returntype' => '\\'.ResultInterface::class,
                     ]
                 )
             );
--- a/test/PhproTest/SoapClient/Unit/CodeGenerator/Assembler/ClientMethodAssemblerTest.php
+++ b/test/PhproTest/SoapClient/Unit/CodeGenerator/Assembler/ClientMethodAssemblerTest.php
@@ -66,7 +66,7 @@ namespace MyNamespace;
 class  extends \Phpro\SoapClient\Client
 {
 
-    public function functionName(\MyTypeNamespace\ParamType \$ParamType) : \MyTypeNamespace\ReturnType
+    public function functionName(\MyTypeNamespace\ParamType \$ParamType) : \Phpro\SoapClient\Type\ResultInterface
     {
         return \$this->call('ParamType', \$ParamType);
     }


veewee:
Hi @GaryJones,

I prefer not to use the interface as a return type. We'll have to investigate how we can fix this in a better way.

GaryJones:
> I prefer not to use the interface as a return type.

Understood, but in this case, it's the only agreed (contracted) certainty (without any other changes).

veewee:
About the result provider issue. might something as this work?

- Add ClientMethodOptions as constructor argument to the ClientMethodAssembler class in which you can specify which class to return.
- When a class has been registered in the options : use that class - otherwise use the automatically generated one.

This way you can customize the way it works by using the ClientMethodMatchesRule and possibly by adding a `ClientMethodMapRule` that works like the `TypeMapRule`.

Then you could do something like this:

```php
return Config::create()
    ->addRule(
        new Rules\ClientMethodMapRule(
            [
                'ClientMethod1' => new Rules\AssembleRule(new Assembler\ClientMethodAssembler(
                        (new ClientMethodAssemblerOptions())->withReturnType(SomeReturnType::class)
                    )
                ),
                'ClientMethod2' => new Rules\AssembleRule(new Assembler\ClientMethodAssembler(
                        (new ClientMethodAssemblerOptions())->withReturnType(ResultInterface::class)
                    )
                )
            ],
            new Rules\AssembleRule(new Assembler\ClientMethodAssembler(new ClientMethodAssemblerOptions()))
        )
    );
```

Or easier to configure:

```php
// This could be in a separate file:
$clientMethodMap = [
    'ClientMethod1' => SomeReturnType::class,
    'ClientMethod2' => ResultInterface::class,
    'ClientMethod3' => ResultInterface::class,
    'ClientMethod4' => ResultInterface::class,
    'ClientMethod5' => ResultInterface::class,
    'ClientMethod6' => ResultInterface::class,
    'ClientMethod7' => ResultInterface::class,
    'ClientMethod8' => ResultInterface::class,
];

return Config::create()
    ->addRule(
        new Rules\ClientMethodMapRule(
            array_map(
                function($methodReturnType) {
                    return new Rules\AssembleRule(new Assembler\ClientMethodAssembler(
                        (new ClientMethodAssemblerOptions())->withReturnType($methodReturnType)
                    );
                },
                $clientMethodMap
            ),
            new Rules\AssembleRule(new Assembler\ClientMethodAssembler(new ClientMethodAssemblerOptions()))
        )
    );

```


What do you think?

GaryJones:
This was never really about having the ability to change what return type was used; it was about the automatic value being incorrect when ResultProviders are present.

Saying that...

I think the `ClientMethodAssemblerOptions` approach can initially be done without the mapping extras.

```php
// Client method created with correct concrete type (which is the bug still to be fixed)
->addRule( new Rules\ClientMethodMatchesRule(
    new Rules\AssembleRule(new Assembler\ClientMethodAssembler(
       (new ClientMethodAssemblerOptions())
    )
    '/^ClientMethod1$/'
) 

// ...or if ClientMethodAssemblerOptions is made optional
->addRule( new Rules\ClientMethodMatchesRule(
    new Rules\AssembleRule(new Assembler\ClientMethodAssembler(),
    '/^ClientMethod1$/'
) 

// ----

->addRule( new Rules\ClientMethodMatchesRule(
    new Rules\AssembleRule(new Assembler\ClientMethodAssembler(
       (new ClientMethodAssemblerOptions())->withReturnType(ResultInterface::class)
    ),
    '/^ClientMethod1$/'
) // Client method returns a different type.
```

In my case, this won't really help, since when I need to return a type that is the interface, it's because I'm also changing the body of the method.

----

We don't need the above to solve #121. My "lazy" solution of defaulting to the interface is perhaps not sufficient here, but the logic should be on getting a better version of `'returntype' => '\\'.$method->getParameterNamespace().'\\'.$method->getReturnType(),` so that it _automatically_ accounts for ResultProviders.

Can we consistently determine what the Result class will be, from a ResultProvider? i.e.:
```
// pseudo-code

$returnType = '\\'.$method->getParameterNamespace().'\\'.$method->getReturnType();
if ( method_exists( `getResult`, $returnType ) {
    $returnType = get_class(new $returnType())->getResult());
}
```

janvernieuwe:
Sounds like a job for https://secure.php.net/manual/en/function.is-subclass-of.php to me ;)

janvernieuwe:
Perhaps something like this might do the trick, as long as you have a return type.
```
<?php

class Bar {}

class Test {
    public function foo(): Bar
    {

    }
}
$function = new ReflectionClass(Test::class);
echo $function->getMethod('foo')->getReturnType();
```

veewee:
Not really, the `ResultProvider` is added by an assembler. We can't assume the types exist yet while creating the client. The only way to determine what the method should return is by explicitly specifying it. The same goes with registering ResultProviders to types, but for types it is easier since it can have the same class for multiple regex matches.

@GaryJones : Agree - the regex can also be used for the same purpose, but is less maintainable. But it is sufficient.

Conclusion: While using ResponseProviders on types and client methods, you always need to explicitly set the return type somehow. Be it by using a regex rule or a map rule.

GaryJones:
Clearly, this needs a bit more thought. Would you be willing to accept this PR, so we at least get a valid return type, even if not quite ideally implemented, and then work on a "better" solution?

> The client generator is mostly used to generate a lot of code which can be customized afterwards. You could add custom assemblers, but I am not sure this has any advantages over generating the client once and adding custom functionallity to it.

There absolutely is value in not having to do manual customisations. The API I happen to be working with has recently pulled a previously public method out of its unversioned API 😱️. Being able to do `composer generate`, and have client, classmap and types be automatically generated correctly, and not then have to customise further, is what I'm after. There are multiple folks working on this package, and to expect them to know how to do any manual customisations after a fresh generation is unreasonable. What I've got working is fantastic (thank you!), so trying to get those fixes upstream to here is the next logical step.

PR #125, opened 2018-01-02
Binary can't find autoload.php

icanhazstring:
This fixes an issue where the binary couldn't find the composer `autoload.php`.
This was caused due to the double checking if existing files.

I removed the old behavior and added a simple list where the autoload.php could be located.

### DIFF
--- a/bin/soap-client
+++ b/bin/soap-client
@@ -1,19 +1,19 @@
 #!/usr/bin/env php
 <?php
 
-define('SOAP_PATH', realpath(__DIR__ . '/..'));
-
-if (is_file($autoload = getcwd() . '/vendor/autoload.php')) {
-    require $autoload;
-} elseif (is_file($autoload = getcwd() . '/../../autoload.php')) {
-    require $autoload;
+for($i=4; $i>=0; --$i) {
+    $autoloadFiles[] = str_pad('/vendor/autoload.php', 20 + ($i * 3), '/..', STR_PAD_LEFT);
 }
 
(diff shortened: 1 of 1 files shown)


icanhazstring:
This issue raised when working with a windows/unix synced dev environment.
Then the binary links from composer will be resolved under `vendor/bin/soap-client` instead of the link `vendor/bin/soap-client => vendor/bin/phpro/soap-client/bin/soap-client`

veewee:
Hi @icanhazstring,

Can you tell me how you run the commands on windows?
The idea of the original code was:

- `getcwd()` if you run from document root in a package
- `__DIR__` if you want to run the client from a global installed  package.
- both if you run from a package dir without the soap-client, which will fall back on the globally installed package

This PR would break those cases.
It seems better to change the autoloading logic to something like this:
https://github.com/phpro/grumphp/blob/master/bin/grumphp#L6-L28

icanhazstring:
Hi @veewee sorry for the late response.

The thing I did was simply check out the repo using my IDE (phpstorm).
This way my source is stored with windows filesystem. I then added a Vagrantfile to simply support using PHP and Composer using a VM.

If I touched the binary file. The link gets destroyed. Then I can't use in inside the VM anymore.
The only way to solve this, is to maybe use the link you provided. I can rewrite it to that one if you want.

veewee:
Feel free te rewrite it.

veewee:
Closing this PR. If this is still an issue, feel free to create a new PR.

PR #128, opened 2018-01-18
Change wrongly referenced interface in docs/client.md

kick-the-bucket:
`MixedArgumentRequestInterface` does not exist and the `MultiArgumentRequest` suggests that it should be `MultiArgumentRequestInterface` instead.


veewee:
Thanks for detecting!

PR #131, opened 2018-02-05
Also Fetches SOAP action from Content-type header

igorsantos07:
The SOAP Action is not always defined in a specific header, so care must be taken when fetching the final action. It might be included on the content-type header as well (quite ugly but.. SOAP).

### DIFF
--- a/src/Phpro/SoapClient/Middleware/WsaMiddleware.php
+++ b/src/Phpro/SoapClient/Middleware/WsaMiddleware.php
@@ -28,7 +28,22 @@ class WsaMiddleware extends Middleware
         $xml = SoapXml::fromStream($request->getBody());
 
         $wsa = new WSASoap($xml->getXmlDocument());
-        $wsa->addAction($request->getHeader('SOAPAction')[0]);
+
+        $actionHeader = $request->getHeader('SOAPAction');
+        if ($actionHeader) {
+            $wsa->addAction($actionHeader[0]);
+        } else {
+            $contentType = $request->getHeader('Content-Type')[0];
+            $parts = explode(';', $contentType);
+            foreach ($parts as $part) {
+                if (strpos($part, 'action=') !== false) {
+                    $actionParts = explode('=', $part);
+                    $action = trim($actionParts[1], '"');
+                    $wsa->addAction($action);
+                }
+            }
+        }
+        
         $wsa->addTo((string) $request->getUri());
         $wsa->addMessageID();
         $wsa->addReplyTo($this->address);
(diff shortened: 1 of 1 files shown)


src/Phpro/SoapClient/Middleware/WsaMiddleware.php
```diff
@@ -28,7 +28,22 @@ public function beforeRequest(callable $handler, RequestInterface $request): Pro
         $xml = SoapXml::fromStream($request->getBody());
 
         $wsa = new WSASoap($xml->getXmlDocument());
-        $wsa->addAction($request->getHeader('SOAPAction')[0]);
+
```
veewee: Can you move this logic to a new private method like `detectSoapAction($request): string`?
This way you can avoid the else structure and throw an exception when no action could be found.

src/Phpro/SoapClient/Middleware/WsaMiddleware.php
```diff
@@ -28,7 +28,22 @@ public function beforeRequest(callable $handler, RequestInterface $request): Pro
...
+        if ($actionHeader) {
+            $wsa->addAction($actionHeader[0]);
+        } else {
+            $contentType = $request->getHeader('Content-Type')[0];
+            $parts = explode(';', $contentType);
+            foreach ($parts as $part) {
+                if (strpos($part, 'action=') !== false) {
+                    $actionParts = explode('=', $part);
+                    $action = trim($actionParts[1], '"');
+                    $wsa->addAction($action);
```
veewee: It is possible that the addAction is triggered multiple times. This can easily be fixed by returning the action in a private method.

src/Phpro/SoapClient/Middleware/WsaMiddleware.php
```diff
@@ -28,7 +28,22 @@ public function beforeRequest(callable $handler, RequestInterface $request): Pro
...
+        $actionHeader = $request->getHeader('SOAPAction');
+        if ($actionHeader) {
+            $wsa->addAction($actionHeader[0]);
+        } else {
+            $contentType = $request->getHeader('Content-Type')[0];
+            $parts = explode(';', $contentType);
+            foreach ($parts as $part) {
+                if (strpos($part, 'action=') !== false) {
+                    $actionParts = explode('=', $part);
+                    $action = trim($actionParts[1], '"');
```
veewee: Also single quotes can be used in the header. You might also want to trim these

igorsantos07:
Attentive peer review. I like that :)

veewee:
The code looks good.
However, we also really need tests for ALL possilble cases to make sure this works. You can find the test cases at `test/PhproTest/SoapClient/Unit/Middleware/WsaMiddlewareTest.php`

PR #132, opened 2018-02-06
Formatting markup on debug info

igorsantos07:
Funny how I wrote this exact code I'm adding (including the array format and keys!) into my generated client, and later on I stumbled upon the original `debugLastSoapRequest()` (except I called mine simply `debug()` haha).

Pushing it up. As that's a debug operation, XML formatting wouldn't be a performance hit.


src/Phpro/SoapClient/Client.php
```diff
@@ -58,15 +58,35 @@ public function applySoapHeaders($soapHeaders): self
      */
     public function debugLastSoapRequest(): array
     {
+        $format = function ($xml) {
```
veewee: Can you move the formatting logic to a separate formatter class to keep the Client class clean?
This way you can also fully test all possible cases.

src/Phpro/SoapClient/Client.php
```diff
@@ -58,15 +58,35 @@ public function applySoapHeaders($soapHeaders): self
...
+            if (!$xml) {
+                return '';
+            }
+
+            $doc = new \DOMDocument('1.0');
+            $doc->formatOutput = true;
+            if (strpos(strtolower($xml), '<html') !== false) {
```
veewee: You can avoid these nested ifs

src/Phpro/SoapClient/Client.php
```diff
@@ -58,15 +58,35 @@ public function applySoapHeaders($soapHeaders): self
...
+                if ($doc->loadHTML($xml)) {
+                    return $doc->saveHTML();
+                }
+            } else {
```
veewee: You can avoid else here

veewee:
Thanks for the PR!
`debug()` might be shorter indeed untill you encounter a "debug" soap action like I did at the time ;)

src/Phpro/SoapClient/Client.php
igorsantos07: I sometimes cannot understand the issue with ifs and elses. Is that to diminish test sub-paths?

src/Phpro/SoapClient/Client.php
igorsantos07: A new class? Or did you mean a private formatter method? That's way simple for a class. That's even the reason with I didn't extract the function from the method.

veewee: A new class is fine. It's simple but SRP :)

igorsantos07: sometimes I wonder how long we go for code organization and strictly following so-called patterns, creating cluttered repositories... but oh well :)

src/Phpro/SoapClient/Client.php
igorsantos07: Read about it a while ago and didn't agree to a couple of items. Re-read again, and I still don't think avoiding `else` makes for more readable code. It actually feels more like "alright, nothing happened, so let's do this by default" when actually a lot of cases it feels more natural to read "this is not what I was expecting, so do that instead".

But that's only my opinion, after all :)

veewee: I understand your point. But at the end, I'll probably be the one digging in somebody else's code to detect some exotic bug :) That's why I like to keep everything as clean as possible.

igorsantos07:
Yeah! I know, that's why I'm working on that even not liking the idea.

After all, there's always something to learn, right?

janvernieuwe:
@veewee Shall we merge this? It looks good to me, did a few small tweaks / synced the PR.

PR #133, opened 2018-02-07
Allowing unencrypted WSSE headers to be used

igorsantos07:


veewee:
Hi @igorsantos07,

Thanks for the PR. Didn't know people were actually using unsigned WSSE.
The PR now contains optional arguments which make it possible to add a private key without a public key which will cause issues.
I would prefer an empty constructor and an explicit `sign($privateKey, $publicKey)` method.


We also really need tests for ALL possilble cases to make sure this works. You can find the test cases at `test/PhproTest/SoapClient/Unit/Middleware/WsaMiddlewareTest.php`

veewee:
Closed due to inactivity

PR #136, opened 2018-02-08
Ensure dir exists before using file_put_contents

janvernieuwe:


src/Phpro/SoapClient/Util/Filesystem.php
```diff
@@ -57,8 +57,9 @@ public function getFileContents(string $path): string
      * @param string $path
      * @param mixed $content
      */
-    public function putFileContents(string $path, $content)
+    public function putFileContents(string $path, $content): void
```
veewee: :void is php >=7.1
Can you remove it?

PR #137, opened 2018-02-08
Add scafolding wizard and consequent generator ouput

veewee:
This new 'wizard' command makes it possible to get somebody who is new to the project up and running in no time.
I also made sure all generators cli tools use SymfonyStyle to format the output. This makes the project more consistent

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,19 @@ Great, you made it so far! Let's find out how this module works:
 $ composer require phpro/soap-client
 ```
 
+## Scafolding Wizard
+
+Since life is too short to read documentation,
+ we've added a scafolding wizard which will get you communicating with your SOAP server in no time!
+All you need to do is:
+
+```sh
+composer require --dev zendframework/zend-code:^3.1.0
+./vendor/bin/soap-client wizard
+```
+
+You can customize the generated code based on the manual installation pages in the next chapter.
+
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -73,6 +73,19 @@ Great, you made it so far! Let's find out how this module works:
...
 ```
 
+## Scafolding Wizard
+
+Since life is too short to read documentation,
+ we've added a scafolding wizard which will get you communicating with your SOAP server in no time!
+All you need to do is:
+
+```sh
+composer require --dev zendframework/zend-code:^3.1.0
```
janvernieuwe: It would be nice if we get a message that zend code should be installed to use this feature when it is not available locally.

src/Phpro/SoapClient/Console/Command/GenerateConfigCommand.php
```diff
@@ -36,20 +37,33 @@ public function __construct(Filesystem $filesystem)
         parent::__construct();
     }
 
-    protected function configure(): void
```
janvernieuwe: No void?

README.md
veewee: Lets keep this for now. Maybe at some point, we can move the code-generation part to a separate repo which requires zend-code.

PR #138, opened 2018-02-08
Always add getters to every generated type

veewee:
This small change adds getters by default - even if you don't add request / response regex.
This makes the client easier to use by default when using the new wizard command.

Sample code:

```php
<?php

return Config::create()
    ->setWsdl('wsdl.xml')
    ->setTypeDestination('src/type')
    ->setTypeNamespace('App\\\\Type')
    ->setClientDestination('src/client')
    ->setClientName('Client')
    ->setClientNamespace('App\\\\Client')
    ->setClassmapDestination('src/classmap')
    ->setClassmapName('Classmap')
    ->setClassmapNamespace('App\\\\Classmap')
    ->addRule(
        new Rules\AssembleRule(new Assembler\GetterAssembler(new Assembler\GetterAssemblerOptions()))
    )
    ->addRule(
        new Rules\TypenameMatchesRule(
            new Rules\MultiRule([
                new Rules\AssembleRule(new Assembler\ConstructorAssembler()),
            ]),
            '/Request$/i'
        )
    )
    ->addRule(
        new Rules\TypenameMatchesRule(
            new Rules\MultiRule([
                new Rules\AssembleRule(new Assembler\ResultAssembler()),
            ]),
            '/Response$/i'
        )
    )
;

```

### DIFF
--- a/src/Phpro/SoapClient/CodeGenerator/ConfigGenerator.php
+++ b/src/Phpro/SoapClient/CodeGenerator/ConfigGenerator.php
@@ -19,9 +19,8 @@ return Config::create()
 BODY;
 
     const RULESET_DEFAULT = <<<RULESET
-->addRule(
-    new Rules\AssembleRule(new Assembler\GetterAssembler(new Assembler\GetterAssemblerOptions()))
-)
+->addRule(new Rules\AssembleRule(new Assembler\GetterAssembler(new Assembler\GetterAssemblerOptions())))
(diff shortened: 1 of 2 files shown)


janvernieuwe:
What do you think about generating setters too when no regex is given?
Code itself lgtm.

veewee:
I've added immutable setters for every generated type. Even though I don't fully like the generated classes, this makes the code fully useable. From that point on, it's the users responsibility to customize the code generation tool.

PR #143, opened 2018-02-23
Validate PHP syntax of generated code

janvernieuwe:
Add extra generator tests
Validate PHP syntax


test/PhproTest/SoapClient/Util/SyntaxChecker.php
```diff
@@ -0,0 +1,33 @@
...
+     * Validate php code in a string
+     *
+     * @param string $phpCode
+     *
+     * @return bool
+     */
+    public static function isValidPHP(string $phpCode): bool
+    {
+        // Remove docblocks
+        $phpCode = preg_replace('/\/\*\*[.\s\*\w\@\$]*\/]*/m', '', $phpCode);
```
janvernieuwe: Not too happy about this hack, but docblocks give false positives :|

PR #154, opened 2018-04-26
Forcing requirement of SOAP and XML extensions

igorsantos07:
Reinstalling project on a new machine made me notice those extensions are required. They should be specified at the composer file :)


janvernieuwe:
Thanks @igorsantos07 !

Issue #124, opened 2017-12-31
Return value of Phpro\SoapClient\Soap\SoapClientFactory::factory() must be an instance of Phpro\SoapClient\Soap\SoapClient, instance of VCR\Util\SoapClient returned


GaryJones:
Related: #84.

When running unit tests, which have php-vcr attached, I'm getting the following error from the first test:

```sh
TypeError: Return value of Phpro\SoapClient\Soap\SoapClientFactory::factory() must be an instance of Phpro\SoapClient\Soap\SoapClient, instance of VCR\Util\SoapClient returned
```

My php-vcr config in my bootstrap:

```php
// Configure and turn on PHP-VCR.
VCR::configure()->setStorage('json');
VCR::configure()->enableLibraryHooks(['soap']);
if (USE_TEST_API) {
    VCR::configure()->setCassettePath('tests/fixtures/vcrTest');
} else {
    VCR::configure()->setCassettePath('tests/fixtures/vcr');
}
VCR::turnOn();
```

If I change the library hook to be `curl`, then the tests all run and pass fine, but nothing gets stored in the php-vcr cassettes (i.e. it doesn't appear to be listening).

I'm using the default Handler (SoapHandle), and soap-client `dev-master`.

I didn't get the issue when I was using soap-client 0.3, so I suspect the addition of strict types / return type declarations since then may be the cause.

Removing the return type from `Phpro\SoapClient\Soap\SoapClientFactory::factory` allows the tests to run (though in my case, the cassette is still blank).

veewee:
That will indeed cause an issue. 
As suggested in #84, I would suggest you to use the curl library hook and use the Http handle instead of the SoapHandle. In the future we will try to move away from ext-soap anyways.
In the meantime we can indeed drop the typehint of the factory.

killua99:
I'm working with the Http\Adapter\Guzzle6\Client and HttPlugWsdlProvider still I do not get the vcr to save the files.

Could you update example tests to run with the current state of the codes? I think I'm missing something that would be small but hard to capture.

killua99:
Well keeping working with this, I'm not sure how to make this project work with php-vcr.

This are some traces.

```Invalid exception returned from Guzzle6
 /www/ac-components/woocommerce-sagex3/vendor/phpro/soap-client/src/Phpro/SoapClient/Exception/WsdlException.php:31
[...]

Issue #84, opened 2017-10-10
Issues with PHP-VCR


kostajh:
I'm working on integrating php-vcr with my project, but running into difficulties. This is the message I get:

```
You can only add handlers if the SoapClientFactory is returning an instance of Phpro\SoapClient\Soap\SoapClient. Got: VCR\Util\SoapClient
 /Users/kostajh/src/example/drupal/sites/all/vendor/phpro/soap-client/src/Phpro/SoapClient/ClientBuilder.php:199
 /Users/kostajh/src/example/drupal/sites/all/modules/custom/example_integration/src/Api/Client/IntegrationClient.php:180
 /Users/kostajh/src/example/drupal/sites/all/modules/custom/example_integration/src/Api/Client/IntegrationClient.php:87
 /Users/kostajh/src/example/drupal/sites/all/modules/custom/example_integration/tests/src/IntegrationTestBase.php:51
 /Users/kostajh/src/example/drupal/sites/all/modules/custom/example_integration/tests/src/Integration/CodeUnitGuestInfoTest.php:22
```

I'm using PHPUnit 6 with the latest 3.0.0 release of php-vcr-listeners project. Line 180 of `IntegrationClient.php` is calling `$clientBuilder->build()` on a client object I've built up.

Any ideas?

veewee:
Hi @kostajh,

The trick is to use the curl library hook instead of the soap hook if you are using the GuzzleClient.

```php
\VCR\VCR::configure()->enableLibraryHooks(['curl_rewrite']);
\VCR\VCR::turnOn();
```

kostajh:
Hmm, so if I set it to `curl_rewrite` I get `PHP Fatal error:  Uncaught InvalidArgumentException: Library hooks don't exist: curl_rewrite`. Based on [the docs](http://php-vcr.github.io/documentation/libraryhooks/) it seems like just specifying `curl` should be enough. I'm wondering if it's an issue with the NTLM authentication on the server I'm interacting with. Debugging it now...

veewee:
Oh yeah, I've must have copied it from an older package version of VCR :) it should do the trick ...

kostajh:
`curl` isn't working though. It seems like the problem is that the NTLM middleware either isn't getting set for the Wsdl provider, or when the client is built. Here's what my code looks like, which works perfectly when not using PHP VCR:

``` php
  $clientFactory = new ClientFactory(GuestInfo::class);
[...]

PR #159, opened 2018-05-17
Fixes Issue #124: Fixes PHPVcr compatibility issues

hniirane:
Fixes an incompatibility with phpvcr due to SoapClientFactory static typing and PHPVcr code overwriting.

Fixes: #124 #84

### DIFF
--- a/src/Phpro/SoapClient/Soap/SoapClientFactory.php
+++ b/src/Phpro/SoapClient/Soap/SoapClientFactory.php
@@ -53,6 +53,6 @@ class SoapClientFactory
 
         $options = array_merge($defaults, $soapOptions);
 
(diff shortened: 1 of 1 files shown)


veewee:
Hi @hniirane,

Thanks for the PR! Can you use an alias instead of a FQN?
This will prevent breaking the code again while using an automatic code reformatter.

hniirane:
>Thanks for the PR! Can you use an alias instead of a FQN? This will prevent breaking the code again while using an automatic code reformatter.

I tested and an alias works just as well. Changed the PR. Would you like me to interactively rebase changes to squash the commits?

veewee:
It's good like this. Thanks!

PR #161, opened 2018-05-25
Code design update

janvernieuwe:
Removed Parameter extend, just used the Zend ParameterGenerator
Use param name defined in the wsdl


veewee:
@janvernieuwe : This is the other way around of what I was trying to say. Is it possible to use simple value objects for the params and convert them to a ParameterGenerator in the assmebler that is generating the actual method parameter?

Also: When using multiple arguments, we need to wrap those arguments with a `MultiArgumentRequest`.

src/Phpro/SoapClient/CodeGenerator/Model/Parameter.php
```diff
@@ -0,0 +1,62 @@
...
+    private $name;
+
+    /**
+     * @var string
+     */
+    private $type;
+
+    /**
+     * Parameter constructor.
+     *
```
veewee: param / return Docblocks in this class are overhead

src/Phpro/SoapClient/CodeGenerator/Assembler/ClientMethodAssembler.php
```diff
@@ -61,4 +62,27 @@ public function assemble(ContextInterface $context): bool
...
+     *
+     * @return ParameterGenerator
+     */
+    private function createParamsFromMethod(ClientMethod $method): ParameterGenerator
+    {
+        $params = $method->getParameters();
+        if (count($params) > 1) {
+            return ParameterGenerator::fromArray(
+                [
+                    'name' => 'multiArgumentRequest',
```
veewee: Maybe we could add docblocks with more information about the MultiArgumentRequest. For example: which parameters / types are available in the multi request argument?

src/Phpro/SoapClient/CodeGenerator/Assembler/ClientMethodAssembler.php
```diff
@@ -61,4 +62,27 @@ public function assemble(ContextInterface $context): bool
...
+                    'type' => MultiArgumentRequest::class,
+                ]
+            );
+        }
+        $param = array_shift($params);
```
veewee: current() works as well and has less overhead

src/Phpro/SoapClient/CodeGenerator/Model/Parameter.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+namespace Phpro\SoapClient\CodeGenerator\Model;
+
+/**
+ * Class Parameter
```
veewee: Class docblocks are overhead

src/Phpro/SoapClient/CodeGenerator/Assembler/ClientMethodAssembler.php
janvernieuwe: Do you have an idea how this docblock should look like?

janvernieuwe:
After this release i'm going to try to add bilge's code, but first release.

PR #175, opened 2018-07-27
Prefix reserved keywords

janvernieuwe:
Prefixes class names that are reserved keywords with Client.
Lowercase version is true for properties.


janvernieuwe:
Suffix with type instead of prefix with client, rgr

Issue #186, opened 2018-09-17
Request does not contain parameters


max21:
|    Q        |   A
|------------ | -----
| Version     | 0.7.6

### Support Question

Each methods does not contain parameters when request it.
For example : 

`$client = \Example\ERP\ERPClientFactory::factory('mywsdl.xsld');
$soapCalcKey = new \Example\ERP\Type\CalcKey('url','id','key');
$response = $client->get_calc_key($soapCalcKey);
var_dump($client->debugLastSoapRequest());exit;`

The debug request call the right method but does not contains parameters url, id and key.

janvernieuwe:
@max21 can you checkout dev-master, and regenrate your types to see if this solves your issue?

veewee:
Hi @max21,

We consider this issue to be fixed. Feel free to create a new issue with additional WSDL + code generation information if you still have any problems.

PR #187, opened 2018-09-18
Don't normalize property casing

### LINKED ISSUE #185: Issues due to property normalizing
I'm trying to integrate against a SOAP service and I'm running into some issues with the property normalizing.

All properties in the WSDL are in UpperCamelCase (e.g. OutErrorCode, OutNumAvail etc.) and are normalized during code generation to lowerCamelCase (e.g. `$outErrorCode`, `$outNumAvail` etc.). So far so good!

However, problems occur when making requests and trying to use the result classes since our friend SoapClient seems to assign values to the properties' UpperCamelCase variant.

Dumping the result (using `symfony/var-dumper`) outputs something along the lines of:

```
FoobarResponse {#101
  -outErrorCode: null
  -outFoobar: null
  +"OutErrorCode": "11"
  +"OutFoobar": ""
}
```

I assume this isn't the intended behaviour? The workaround right now is to manually change all type properties to the correct case. My assumption is that SoapClient is unaware of the `OutErrorCode -> outErrorCode` transformation, but I'm not sure it's possible to make it aware either.

Thanks!

janvernieuwe:
Fixes #185

### DIFF
--- a/spec/Phpro/SoapClient/CodeGenerator/Util/NormalizerSpec.php
+++ b/spec/Phpro/SoapClient/CodeGenerator/Util/NormalizerSpec.php
@@ -30,15 +30,21 @@ class NormalizerSpec extends ObjectBehavior
     {
         $this->normalizeClassname('myType')->shouldReturn('MyType');
         $this->normalizeClassname('final')->shouldReturn('FinalType');
+        $this->normalizeClassname('Final')->shouldReturn('FinalType');
+        $this->normalizeClassname('UpperCased')->shouldReturn('UpperCased');
         $this->normalizeClassname('my-./*type_123')->shouldReturn('MyType123');
+        $this->normalizeClassname('my-./final*type_123')->shouldReturn('MyFinalType123');
     }
 
     function it_noramizes_properties()
     {
         $this->normalizeProperty('prop1')->shouldReturn('prop1');
-        $this->normalizeProperty('final')->shouldReturn('finalType');
+        $this->normalizeProperty('final')->shouldReturn('final');
+        $this->normalizeProperty('Final')->shouldReturn('Final');
+        $this->normalizeProperty('UpperCased')->shouldReturn('UpperCased');
         $this->normalizeProperty('my-./*prop_123')->shouldReturn('myProp_123');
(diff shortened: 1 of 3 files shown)


spec/Phpro/SoapClient/CodeGenerator/Util/NormalizerSpec.php
```diff
@@ -30,14 +30,21 @@ function it_can_normalize_classnames()
...
+        $this->normalizeClassname('UpperCased')->shouldReturn('UpperCased');
         $this->normalizeClassname('my-./*type_123')->shouldReturn('MyType123');
+        $this->normalizeClassname('my-./final*type_123')->shouldReturn('MyFinalType123');
     }
 
     function it_noramizes_properties()
     {
         $this->normalizeProperty('prop1')->shouldReturn('prop1');
-        $this->normalizeProperty('final')->shouldReturn('finalType');
+        $this->normalizeProperty('final')->shouldReturn('final');
```
veewee: This is perfectly valid php:
```php
class X {
    private $final;
    private $global;
    private $if;
}
```
https://3v4l.org/GW5aN

If you transform those keyworded properties, ext-soap won't be able to link the xml to the objects.

src/Phpro/SoapClient/CodeGenerator/Util/Normalizer.php
```diff
@@ -110,10 +110,9 @@ private static function normalizeReservedKeywords(string $name, $ucfirst = true)
         if (!\in_array(strtolower($name), self::$reservedKeywords, true)) {
             return $name;
         }
-        $name = ucfirst($name);
         $name .= 'Type';
 
-        return $ucfirst ? ucfirst($name) : lcfirst($name);
+        return $ucfirst ? ucfirst($name) : $name;
```
veewee: Maybe better to drop the ucfirst param and wrap the function where you need it? Seems easier to follow.

janvernieuwe: Agreed, was thinking about it yesterday.

janvernieuwe:
Maybe this can fix #186 too, but there is not enough information in that issue to be sure of that.

PR #192, opened 2018-09-28
Normalize class names in typehints when generating code

### LINKED ISSUE #191: Entity names with underscores are treated inconsistently
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | unknown
| Version     | master

#### Summary

SOAP entities containing underscores are treated inconsistently in generated PHP code.

#### Current behavior

When generating PHP code, a SOAP entity with an underscore in the name will keep the underscore in some generated code, but will lose it in other generated code. The worst result of this is that the method calls generated using `generate:client` will have typehints that don't match the actual class names, causing them to be unusable.

#### How to reproduce

Create a config file with the following contents:

```
<?php

return Config::create()
    ->setWsdl('https://api.smdservers.net/CCWs_3.5/CallCenterWs.asmx?WSDL')
    ->setTypeDestination('src\Type')
    ->setTypeNamespace('Example')
    ->setClientDestination('src')
    ->setClientName('ExampleClient')
    ->setClientNamespace('Example')
    ->setClassMapDestination('src')
    ->setClassMapName('ExampleClassmap')
    ->setClassMapNamespace('Example')
    ->addRule(
        new Rules\AssembleRule(
            new Assembler\GetterAssembler(
                new Assembler\GetterAssemblerOptions()
            )
        )
    )
;
```

Run these commands:

```
vendor/bin/soap-client generate:classmap --config=config/soap-client.php
vendor/bin/soap-client generate:client --config=config/soap-client.php
vendor/bin/soap-client generate:types --config=config/soap-client.php
```

Using the `TenantUpdate_NationalAccount` entity as an example (and the associated Response and Result entities), we can see the following differences:

* The name of the result class in src/Type is `TenantUpdateNationalAccountResult`
* The member name and getter in `TenantUpdateNationalAccountResponse` uses `TenantUpdate_NationalAccountResult` instead
* The function in `CallCenterClient` has the name `tenantUpdate_NationalAccount` and has an invalid typehint (`TenantUpdate_NationalAccount`) and return type (`TenantUpdate_NationalAccountResponse`)

#### Expected behavior

The entity name is treated the same way in all generated code: it either keeps the underscore, or loses it.

clamburger:
Resolves #191.

This change takes a lighter approach that doesn't change the public API - instead, it only fixes the references to class names in docblocks and typehints that were previously broken.

### DIFF
--- a/src/Phpro/SoapClient/CodeGenerator/Model/Property.php
+++ b/src/Phpro/SoapClient/CodeGenerator/Model/Property.php
@@ -57,7 +57,7 @@ class Property
             return $this->type;
         }
 
-        return '\\'.$this->namespace.'\\'.$this->type;
+        return '\\'.$this->namespace.'\\'.Normalizer::normalizeClassname($this->type);
     }
 
     /**
--- a/src/Phpro/SoapClient/CodeGenerator/Parser/FunctionStringParser.php
+++ b/src/Phpro/SoapClient/CodeGenerator/Parser/FunctionStringParser.php
@@ -42,6 +42,7 @@ class FunctionStringParser
         foreach ($properties as $property) {
             list($type, $name) = explode(' ', trim($property));
             $name = Normalizer::normalizeProperty($name);
+            $type = Normalizer::normalizeClassname($type);
             $parameters[] = new Parameter($name, $this->parameterNamespace.'\\'.$type);
         }
 
--- a/test/PhproTest/SoapClient/Unit/CodeGenerator/Assembler/GetterAssemblerTest.php
+++ b/test/PhproTest/SoapClient/Unit/CodeGenerator/Assembler/GetterAssemblerTest.php
@@ -154,6 +154,40 @@ CODE;
     }
 
     /**
+     * @test
+     */
+    function it_assembles_with_normalised_class_name()
+    {
+        $options = (new GetterAssemblerOptions())->withReturnType();
+        $assembler = new GetterAssembler($options);
+
+        $context = $this->createContext('prop4');
+        $assembler->assemble($context);
+
+        $code = $context->getClass()->generate();
+        $expected = <<<CODE
+namespace MyNamespace;
+
+class MyType
+{
+
+    /**
+     * @return \\ns1\\MyResponse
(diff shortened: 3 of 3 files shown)


clamburger:
I've realised that this just change now causes a break in a different place: the underscore ends up being removed from the method name in `$this->call`. This is because ClientMethodAssembler uses the same parameters to fill in both the method parameters, the docblock, and the method body (and we want them to be different, with the underscore only included in the body). At this stage I'm not sure of an elegant way to solve this problem.

joshhornby:
Happy to help look at this @clamburger, can you add a failing test with proves the comment above?

joshhornby:
An example of a failing test which describes what @clamburger commented above:

```
   /**
     * @test
     */
    function it_assembles_a_method_with_underscore_param_type()
    {
        $assembler = new ClientMethodAssembler();
        $class = new ClassGenerator();
        $class->setNamespaceName('Vendor\\MyNamespace');
        $method = ClientMethod::createFromExtSoapFunctionString(
            'ReturnType function_name(param_type $param)',
            'Vendor\\MyTypeNamespace'
        );

        $context = new ClientMethodContext($class, $method);

        $code = $context->getClass()->generate();
        $expected = <<<CODE
namespace Vendor\MyNamespace;

use Phpro\SoapClient\Type\RequestInterface;
use Phpro\SoapClient\Type\ResultInterface;
use Vendor\MyTypeNamespace;
use Phpro\SoapClient\Exception\SoapException;

class  extends \Phpro\SoapClient\Client
{

    /**
     * @param RequestInterface|MyTypeNamespace\ParamType \$param
     * @return ResultInterface|MyTypeNamespace\ReturnType
     * @throws SoapException
     */
    public function functionName(\Vendor\MyTypeNamespace\ParamType \$param) : \Vendor\MyTypeNamespace\ReturnType
    {
        return \$this->call('param_type', \$param);
    }


}

CODE;

        $this->assertEquals($expected, $code);
    }
```

veewee:
Thanks for the example @joshhornby ! I'll close this PR in favour of #196.

Issue #193, opened 2018-10-06
ConstructorAssembler & Type Hints


joshhornby:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | yes

#### Summary

Would it be possible to use type hints in constructors? Take the following example:

```
    /**
     * Constructor
     *
     * @var string $prop1
     * @var int $prop2
     */
    public function __construct($prop1, $prop2)
    {
        $this->prop1 = $prop1;
        $this->prop2 = $prop2;
    }
```

Would become:

```
    public function __construct(string $prop1, int $prop2)
    {
        $this->prop1 = $prop1;
        $this->prop2 = $prop2;
    }
```

IMO is has two purposes, improved type hinting and also removal of the doc block as the function now contains all the information.

PR #194, opened 2018-10-06
Added ConstructorAssemblerOptions

joshhornby:
Added ConstructorAssemblerOptions so we can enable type hinting for constructors. This is a BC break so we need a new version cutting.

Resolves https://github.com/phpro/soap-client/issues/193

## Todo

- [x] Update Docs

### DIFF
--- a/src/Phpro/SoapClient/CodeGenerator/Assembler/ConstructorAssembler.php
+++ b/src/Phpro/SoapClient/CodeGenerator/Assembler/ConstructorAssembler.php
@@ -82,13 +82,19 @@ class ConstructorAssembler implements AssemblerInterface
             $constructor->setParameter(array_merge([
                 'name' => $property->getName(),
             ], $withTypeHints));
-            $docblock->setTag([
-                'name' => 'var',
-                'description' => sprintf('%s $%s', $property->getType(), $property->getName())
-            ]);
+
+            if ($this->options->useDocBlocks()) {
+                $docblock->setTag([
+                    'name' => 'var',
+                    'description' => sprintf('%s $%s', $property->getType(), $property->getName())
+                ]);
+            }
         }
 
-        $constructor->setDocBlock($docblock);
+        if ($this->options->useDocBlocks()) {
+            $constructor->setDocBlock($docblock);
+        }
+
         $constructor->setBody(implode($constructor::LINE_FEED, $body));
 
         return $constructor;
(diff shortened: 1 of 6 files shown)


test/PhproTest/SoapClient/Unit/CodeGenerator/Assembler/ConstructorAssemblerTest.php
```diff
@@ -64,6 +65,42 @@ public function __construct(\$prop1, \$prop2)
...
+namespace MyNamespace;
+
+class MyType
+{
+
+    /**
+     * Constructor
+     *
+     * @var string \$prop1
+     * @var int \$prop2
```
joshhornby: Do we still want to include doc blocks if we are using type hints? For me they add no value when using type hints.

veewee: You could add a `docBlocks` property to the constructor options. This way the end-user can choose how the code is generated.

src/Phpro/SoapClient/CodeGenerator/Assembler/ConstructorAssemblerOptions.php
```diff
@@ -0,0 +1,47 @@
...
+ * Class ConstructorAssemblerOptions
+ *
+ * @package Phpro\SoapClient\CodeGenerator\Assembler
+ */
+class ConstructorAssemblerOptions
+{
+    /**
+     * @var bool
+     */
+    private $withTypeHints = false;
```
veewee: Property should be named `typeHints`. (Remove the with prefix)

src/Phpro/SoapClient/CodeGenerator/Assembler/ConstructorAssembler.php
```diff
@@ -16,6 +16,21 @@
...
+     * @var ConstructorAssemblerOptions
+     */
+    private $options;
+
+    /**
+     * ConstructorAssembler constructor.
+     *
+     * @param ConstructorAssemblerOptions $options
+     */
+    public function __construct(ConstructorAssemblerOptions $options)
```
veewee: This means that the `src/Phpro/SoapClient/CodeGenerator/ConfigGenerator.php` should also be changed.

joshhornby:
Hi @veewee 

Thanks for the review, I've made the requested changes and updated the docs.

Thanks

veewee:
Looks good, thanks for the work!

src/Phpro/SoapClient/CodeGenerator/Assembler/ConstructorAssembler.php
jagers: What about making this an optional param to maintain backwards compatibility? All my existing config files broke because they were written prior to this change getting merged in

PR #196, opened 2018-10-12
Normalization fixes

### LINKED ISSUE #191: Entity names with underscores are treated inconsistently

veewee:
Fixes #192
Fixes #191

This PR contains:
- general type normalization written by @clamburger.
- improved client method naming suggested by @joshhornby.
- bugfix: use function name based on soap information instead of based on type.
- Rename valid soap method names that aren't valid in PHP.

### DIFF
--- a/src/Phpro/SoapClient/CodeGenerator/Util/Normalizer.php
+++ b/src/Phpro/SoapClient/CodeGenerator/Util/Normalizer.php
@@ -101,16 +101,17 @@ class Normalizer
 
     /**
      * @param string $name
+     * @param string $suffix
      *
      * @return string
      */
-    private static function normalizeReservedKeywords(string $name): string
+    private static function normalizeReservedKeywords(string $name, string $suffix): string
     {
         if (!\in_array(strtolower($name), self::$reservedKeywords, true)) {
             return $name;
         }
 
-        return $name.'Type';
+        return $name.$suffix;
     }
 
     /**
@@ -142,13 +143,32 @@ class Normalizer
     }
 
     /**
+     * @param string $method
+     *
+     * @return string
+     */
+    public static function normalizeMethodName(string $method): string
+    {
+        // Methods cant start with a number in PHP - move it after text
+        $method = preg_replace('{^([0-9]*)(.*)}', '$2$1', $method);
+        if (is_numeric($method)) {
+            $method = 'call' . $method;
+        }
+
+        // Methods cant be named after reserved keywords.
+        $method = self::normalizeReservedKeywords($method, 'Call');
+
+        return lcfirst(self::camelCase($method, '{[^a-z0-9]+}i'));
+    }
+
+    /**
      * @param string $name
      *
      * @return string
      */
     public static function normalizeClassname(string $name): string
     {
-        $name = self::normalizeReservedKeywords($name);
+        $name = self::normalizeReservedKeywords($name, 'Type');
 
         return ucfirst(self::camelCase($name, '{[^a-z0-9]+}i'));
     }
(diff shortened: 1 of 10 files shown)


joshhornby:
Just tested this, and works perfectly!

joshhornby:
@veewee Just tested this and not sure this is working quite yet. It fixes the docblocks, method name and type hint but not the return type. Please see the example below:

```
    /**
     * @param RequestInterface|Type\get_application_note $parameters
     * @param RequestInterface|Type\GetApplicationNote $parameters
     * @return ResultInterface|Type\get_application_noteResponse
     * @throws SoapException
     */
    public function get_application_note(\Foo\Bar\Type\get_application_note $parameters) : \Foo\Bar\Type\get_application_noteResponse
    public function getApplicationNote(\Foo\Bar\Type\GetApplicationNote $parameters) : \Foo\Bar\Type\get_application_noteResponse
    {
        return $this->call('get_application_note', $parameters);
    }
```

Notice `\Foo\Bar\Type\get_application_noteResponse` is still incorrect and should be `\Foo\Bar\Type\GetApplicationNoteResponse`

veewee:
@joshhornby, Thanks for testing it out. I've fixed the remark.

joshhornby:
@veewee Just re-tested and can confirm this works as expected. Great work.

src/Phpro/SoapClient/CodeGenerator/Util/Normalizer.php
```diff
@@ -101,16 +101,17 @@ class Normalizer
 
     /**
      * @param string $name
+     * @param string $suffix
      *
      * @return string
      */
-    private static function normalizeReservedKeywords(string $name): string
+    private static function normalizeReservedKeywords(string $name, string $suffix): string
```
janvernieuwe: Wouldn't it be better to add the suffix where this is called?

PR #200, opened 2018-10-30
Default to WSDL_CACHE_DISK caching strategy

mlamm:
The soap caching strategy WSDL_CACHE_MEMORY (WSDL_CACHE_BOTH) caused random segmentation faults while using this library, after debugging the issue we found that WSDL_CACHE_DISK is much more reliable and makes the crashes go away.

So Disk-Caching should be the default in here.

### DIFF
--- a/src/Phpro/SoapClient/Soap/SoapClientFactory.php
+++ b/src/Phpro/SoapClient/Soap/SoapClientFactory.php
@@ -47,7 +47,7 @@ class SoapClientFactory
             'trace' => true,
             'exceptions' => true,
             'keep_alive' => true,
-            'cache_wsdl' => WSDL_CACHE_BOTH,
+            'cache_wsdl' => WSDL_CACHE_DISK,
             'features' => SOAP_SINGLE_ELEMENT_ARRAYS,
             'classmap' => $this->classMap->toSoapClassMap(),
             'typemap' => $this->typeConverters->toSoapTypeMap(),


veewee:
Hi @mlamm,

Thanks for the pr!

Can you tell me more about the segfault?
I had similar issues which could be solved by increasing the `soap.wsdl_cache_limit` ini setting:  http://php.net/manual/en/soap.configuration.php#ini.soap.wsdl-cache-limit

More info about the bug: https://bugs.php.net/bug.php?id=50722

As you state: caching to disk is better but I would prefer no caching by default. This way you can create your own cache file and store it in a faster caching backend like redis.
Iirc: the cache file is just a copy of the wsdl with all it's external xsd files downloaded into this file. But I have to reinvestigate that statement that be sure.

mlamm:
Hey veewee, thanks for taking a look at this.

Unfortunately we weren't able to track down the exact line and file of the php-code, i talked to the guys in #php on freenode and they suggested staying away from memory caching. 

I spinned up some debug container to debug the crash, however, I always did run into a failing assertion on this line https://github.com/php/php-src/blame/php-7.2.11/ext/soap/php_sdl.c#L3024, that code is only used for memory caching, so I just turned it off completely. The failing assert is not necessarily the crash location, I never figured out where it's actually crashing since it also happened randomly.

I just wanted to have my founding flowing back to the soap lib here, to avoid other people running into the very same issue. However, changing default values can be tricky, since some users might rely on them.
Another thing that makes it confusing, the lib overwrites those flags always, so settings in php.ini or ini_set() are silently ignored.

veewee:
I've set the default to disk caching in the new version (#197).
Closing this one for now.

Issue #198, opened 2018-10-24
Error : Wrong parameters for Phpro\SoapClient\Exception\SoapException


jagers:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes/no
| Version     | 0.7.8
| PHP Version | 7.1

#### Summary

When`Phpro\SoapClient\Client::call()` catches an Exception and tries to wrap it in a SoapException, if  the original exception has a non-integer $code value, the `SoapException::fromThrowable()` instantiation will fail with the error `Error : Wrong parameters for Phpro\SoapClient\Exception\SoapException([string $message [, long $code [, Throwable $previous = NULL]]])`

Per [PHP Exception::getCode docs ](http://php.net/manual/en/exception.getcode.php), it is valid for getCode to return a string:
> Returns the exception code as integer in Exception but possibly as other type in Exception descendants (for example as string in PDOException).

#### Current behavior

Execution of the code halts with the error `Error : Wrong parameters for Phpro\SoapClient\Exception\SoapException`

#### How to reproduce

I ran into this issue due to a PDOException that had an error code of `hy000` set.
In my environment, I have a middleware that uses `beforeRequest()` to attempt to save the `$request` to the database via Laravel's Eloquent ORM. If a database query ever throws a PDOException, that PDOException has a non-integer code, so when `SoapException::fromThrowable()`  tries to create self using `$exception->getCode()`, PHP throws the error `PHP Error:  Wrong parameters for Phpro/SoapClient/Exception/SoapException([string $message [, long $code [, Throwable $previous = NULL]]])`

The simplest way to reproduce this is to imitate the behavior of the PDOException by creating a custom exception and throwing it in a middleware that is registered in a client:

```
class MyPDOException extends \PDOException {

    public function __construct(string $message = "", int $code = 0, \Throwable $previous = null)
    {
        parent::__construct($message, $code, $previous);
        $this->code = 'hy000';
    }
}
class MyMiddleware extends \Phpro\SoapClient\Middleware\Middleware
{
    public function getName(): string
    {
        return static::class;
    }
    public function beforeRequest(callable $next, RequestInterface $request): Promise
    {
        throw new MyPDOException("Some MySQL Error");
    }
}
```

veewee:
Hi @jagers,

[...]

PR #201, opened 2018-10-31
Fix issue 198 by setting exception code after parent constructor

jagers:
This fix for issue #198 preserves the $code of the original exception


janvernieuwe:
I'm not too sure about this.
You can always use the previous exception to get the code.

janvernieuwe:
Fixed in https://github.com/phpro/soap-client/pull/199

Issue #218, opened 2019-01-29
`xsd:simpleType` properties are resolved to non existing classes during code generation


ademarco:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 0.7.11

#### Summary

When generating code `xsd:simpleType`properties are resolved to a non existing type class instead of being resolved to their actual base type.

#### Current behavior

Take the current XSD snippet:

```xml
    <xsd:simpleType name="languageCode">
        <xsd:restriction base="xsd:string">
            <xsd:enumeration value="ML"/>
            <xsd:enumeration value="EN"/>
            <xsd:enumeration value="FR"/>
            <xsd:enumeration value="DE"/>
        </xsd:restriction>
    </xsd:simpleType>
```

This will generate the following property in the type classes that use that property:

```php
namespace MyLibrary\Service\Type;

class Document
{
    /**
     * @var MyLibrary\Service\Type\LanguageCode
     */
    protected $language;
}
```

The `MyLibrary\Service\Type\LanguageCode` class is not generated as it's not a complex type and only complex types are collected for generation in `SoapClient::getSoapTypes()`.

#### How to reproduce

You can reproduce by checking out [this commit](https://github.com/openeuropa/epoetry-client/commit/a4a708043e6c8212ae99190f9c7edc05a37c024e) of our library and run a code generation.

You can see also the bug at [this line](https://github.com/openeuropa/epoetry-client/commit/a4a708043e6c8212ae99190f9c7edc05a37c024e#diff-2a0a2d15bfbb4fe8055397f64a512023R11), for example.

#### Expected behavior

The expected behavior is to fallback to the primitive type the element refers to in its XSD definition, meaning, given:

```xml
    <xsd:simpleType name="languageCode">
        <xsd:restriction base="xsd:string">
        ...
```

We should get:

```php
namespace MyLibrary\Service\Type;

class Document
{
    /**
     * @var string
     */
    protected $language;
}
```

veewee:
Thanks for reporting.

The problem with enum types is that we don't get the base type from ext-soap. With other words: we won't be able to provide that language is a string and not an int as long as we are relying on ext-soap.

I agree that we could add a mapping to the normalizer for other main simpleTypes like `base64Binary` etc.

So about your question in general: It is not something we can fix at this point.

ademarco:
About:

> we won't be able to provide that language is a string and not an int as long as we are relying on ext-soap.

[...]

PR #219, opened 2019-01-29
`xsd:simpleType` properties are resolved to non existing classes during code generation

ademarco:
Ref. #218

### DIFF
--- a/src/Phpro/SoapClient/Soap/SoapClient.php
+++ b/src/Phpro/SoapClient/Soap/SoapClient.php
@@ -114,12 +114,8 @@ class SoapClient extends \SoapClient
                     continue;
                 }
                 preg_match('/\s* (.*) (.*);/', $line, $matches);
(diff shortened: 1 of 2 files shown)


src/Phpro/SoapClient/Soap/SoapClient.php
```diff
@@ -91,6 +91,16 @@ public function getSoapTypes(): array
         }
 
         $soapTypes = $this->__getTypes();
+
+        $simpleTypes = [
+            'base64Binary' => 'string',
```
veewee: Shouldn't we just add this to the Normalization::$normalizations  array?
That way it is fixed in code generation and the result of the types stays as close to the actual wsdl as possible.
That is the main idea of the rewritten parser for the next release anyways (see #197).

veewee:
Will be fixed in #197

PR #225, opened 2019-03-04
Don't wrap doc blocks (Fixes #221)

### LINKED ISSUE #221: DocBlockGenerator allow to set wordrwrap
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

It's wierd when docblock is over multiple rows:
```
     /**
      * @param 
      * SomeLongNamespace\Type\Blaa
      * $paramName
      *
```

Id like to be able to turn off this, also I'm running phpcbf after this is generated which automatically imports all those

MidnightDesign:


janvernieuwe:
Hi @MidnightDesign ,

It looks good, thanks for your PR, but we're thinking it might be an even better idea to have a factory that sets some defaults for our project so that this is centralized (and easier to test).

This factory could have a fromArray method where you can pass the other options into and returns the docblock generator with default options set for the project.

This way we don't need to repeat this everywhere.

MidnightDesign:
That's what I was thinking as well, I just wanted to keep the PR as simple as possible. If you want I can implement that.

janvernieuwe:
If you would, that would be awesome.
Thanks in advance!

src/Phpro/SoapClient/CodeGenerator/Assembler/DocBlockGeneratorFactory.php
```diff
@@ -0,0 +1,15 @@
+<?php declare(strict_types=1);
+
+namespace Phpro\SoapClient\CodeGenerator\Assembler;
```
veewee: Maybe better to place this in a separate namespace instead of the assemblers namespace.
I was thinking about something like: `Phpro\SoapClient\CodeGenerator\ZendCodeFactory`
(Naming is hard: so if you have a better idea, feel free to suggest! )

veewee:
Thanks @MidnightDesign !

PR #240, opened 2019-05-20
Fixes #239: Use of $this->soapClient in debugLastSoapRequest results in "Undefined property" error

### LINKED ISSUE #239: Use of $this->soapClient in debugLastSoapRequest results in "Undefined property" error
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 1.0.2

#### Summary

There is a regression in [Client::debugLastSoapRequest](https://github.com/phpro/soap-client/blob/1.0.2/src/Phpro/SoapClient/Client.php#L44) introduced in [this commit](https://github.com/phpro/soap-client/commit/5f3cc796b1f1b7c3985627d1a6e125c2b1b99284) and released in 1.0.2.

We can see the original commit [here](https://github.com/dxvargas/soap-client/commit/753ca87e18e94788be7505669e6fbda485fad838), the change was applied to `$this->soapClient` but in the more recent versions what we had there was `$lastRequestInfo` which uses `$this->engine` instead of `$this->soapClient`. Now we can see that `$lastRequestInfo` is still there but it's not used.

It all looks like the conflict mentioned in [this commit](https://github.com/phpro/soap-client/commit/e37513d786e6ea96111cd239bdc5e67217e7e3c2) was not well resolved.

#### Current behavior

With version 1.0.1 unit tests were passing, with version 1.0.2 I have this error:
`Undefined property: RequestClient::$soapClient`

#### How to reproduce

When calling a SOAP method with the client, something like:
`
$client->soapMethod($request);
$debug = $client->debugLastSoapRequest()['request'];
`
I get the error:
`Undefined property: RequestClient::$soapClient`

This can be seen here:
https://drone.fpfis.eu/openeuropa/epoetry-client/257/13
which originated from [this commit|https://github.com/openeuropa/epoetry-client/pull/18/commits/4e4e859ec6b380a45f7522a763554e2e07ae4a17] here:
https://github.com/openeuropa/epoetry-client/pull/18

I hope, if needed, it's easy to put up this environment and reproduce the error there.

#### Expected behavior

No error as before with 1.0.1.

dxvargas:
Avoid using soapClient in debugLastSoapRequest method of the Client.

### DIFF
--- a/src/Phpro/SoapClient/Client.php
+++ b/src/Phpro/SoapClient/Client.php
@@ -46,12 +46,12 @@ class Client implements ClientInterface
         $lastRequestInfo = $this->engine->collectLastRequestInfo();
         return [
             'request' => [
-                'headers' => trim($this->soapClient->__getLastRequestHeaders()),
-                'body'    => XmlFormatter::format($this->soapClient->__getLastRequest()),
+                'headers' => trim($lastRequestInfo->getLastRequestHeaders()),
+                'body'    => XmlFormatter::format($lastRequestInfo->getLastRequest()),
             ],
             'response' => [
-                'headers' => trim($this->soapClient->__getLastResponseHeaders()),
-                'body'    => XmlFormatter::format($this->soapClient->__getLastResponse()),
+                'headers' => trim($lastRequestInfo->getLastResponseHeaders()),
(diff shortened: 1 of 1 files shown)


veewee:
Thanks for the PR!

PR #244, opened 2019-06-07
Add PHPStan for additional bug detection

### LINKED ISSUE #241: Add integration tests for the Client class
### Improvement Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

Currently there are no real tests for the base `Phpro\SoapClient\Client` class.
This resulted in #239.

We should add integration tests to make sure both the debug method, the event system and the call method does what it does.

janvernieuwe:
PHPStan found the issue mentioned, and a missing interface as well, Interop is deprecated and moved to a psr lib.
Suppressed one false positive due to SoapClient reflection weirdness in the phpstan.neon config file.

### DIFF
--- a/grumphp.yml
+++ b/grumphp.yml
@@ -13,3 +13,8 @@ parameters:
         phpspec: ~
         phpunit: ~
         phplint: ~
+        phpstan:
+            configuration: "phpstan.neon"
+            ignore_patterns:
+                - "spec/"
+                - "vendor/"
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,0 +1,3 @@
+parameters:
+  ignoreErrors:
+    - '#Call to an undefined static method SoapClient::__construct().#'
--- a/composer.json
+++ b/composer.json
@@ -39,12 +39,14 @@
     "phpro/grumphp": "~0.15",
     "phpspec/phpspec": "~5.1",
     "phpspec/prophecy": "~1.8",
+    "phpstan/phpstan": "^0.11.8",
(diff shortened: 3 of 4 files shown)


phpstan.neon
```diff
@@ -0,0 +1,3 @@
+parameters:
+  ignoreErrors:
+    - '#Call to an undefined static method SoapClient::__construct().#'
```
veewee: looks like invalid configuration or possible a bug in phpstan? Or does it expect PHP4 style constructors? :D
Not to fond of adding these kind of lines, but if thats the only solution ....

grumphp.yml
```diff
@@ -12,4 +12,9 @@ parameters:
                 - "\\)\\s?:\\s?void"
         phpspec: ~
         phpunit: ~
-        phplint: ~
\ No newline at end of file
+        phplint: ~
+        phpstan:
```
veewee: This runs phpstan at level 1.
Looks like level 7 detects 92 issues.

phpstan.neon
janvernieuwe: I looked it up, and it has something to to with reflection not working properly on the soap client class.

grumphp.yml
janvernieuwe: Yeah, seems like some interfaces need some methods to be defined still, and some other assorted issues. I'll look at them asap.

veewee:
Merging this one already. Feel free to open up a new PR for the next level stuff.

Issue #246, opened 2019-06-10
`Phpro\SoapClient\Client::debugLastSoapRequest()` returns empty `response.body`


hshhhhh:
Hey, I'm not brightest person on the planet, so excuse me for lack of explanation.

At first I want to say thanks for this client -- it really simplify life and makes soap not so awful thing.

Lately I updated to version `1.0.3` and found that `Phpro\SoapClient\Client::debugLastSoapRequest()` returns empty `response.body`. I would like to store response for debug purposes so I tried to find what's wrong.

At first I found:
- issue https://github.com/phpro/soap-client/issues/239 
- pull requesst: https://github.com/phpro/soap-client/commit/24e4d52ce3f53e7bed23c5f2a6bf89febc14cb72

I do not really understand why it happens, but it happens there: https://github.com/phpro/soap-client/blob/master/src/Phpro/SoapClient/Soap/Engine/Engine.php#L36-L42

When you receive response your `$lastRequestInfo->getLastResponse()` still contains empty response body and fills it only on next line.

```
        $response = $this->handler->request($request);
        return $this->driver->decode($method, $response);
```

I tried to check original sources of soap client and as I can understand it `$client->__last_response` should be filled when you call `parent::__doRequest()`, but for some reason it dosn't: https://github.com/php/php-src/blob/PHP-7.1.27/ext/soap/soap.c#L2590

It definitely happens because you redefined function `__doRequest()`, but I do not get why `AbusedClient::doActualRequest() > parent::__doRequeest()` does'nt fix it.

Please, check PR: https://github.com/phpro/soap-client/pull/247

veewee:
Thanks for the detailed information and the PR!

PR #247, opened 2019-06-10
`Phpro\SoapClient\Client::debugLastSoapRequest()` returns empty `response.body`

hshhhhh:
`Phpro\SoapClient\Soap\Driver\ExtSoap\AbusedClient::doActualRequest()` doesn't set `$this->__last_response` so `Phpro\SoapClient\Client::debugLastSoapRequest()` returns empty `response.body`

https://github.com/phpro/soap-client/issues/246


veewee:
Thanks for the PR. Looks like it indeed solves the issue.
Can you provide a testcase in `\PhproTest\SoapClient\Integration\Soap\Driver\ExtSoap\AbusedClientTest` that covers the `doActualRequest()` method and verifies the request and response methods? 
That way we can be sure we won't break it again in the future.

hshhhhh:
Sure, I will add tests.

PR #248, opened 2019-06-14
Feature: Add ReturnType for Setters

icanhazstring:
Hi. Since all setters are fluent by default. I have just added a valid returntype to it.
I need to add the FQN because `zend-code` will always add a trailing `\` in front. Otherwise I would have taken `self` as a simple approach.


icanhazstring:
@veewee thanks. Sure will see to it.

icanhazstring:
Nice. Thanks! :)

PR #251, opened 2019-06-28
PHPStan level 4

janvernieuwe:
Going up some levels in PHPStan.

New ignores explained:
```
'#Call to an undefined method Phpro\\.*Interface.*$#'
```
The issue we discussed in the chat about the interfaces, hopefully we can fix this in php 7.5

```
'#Invalid type Http\\Client\\Exception to throw.#' 
```
It's an interface that is not throwable, I'll allow it.

More levels next time I get to work on this.

### DIFF
--- a/grumphp.yml
+++ b/grumphp.yml
@@ -1,4 +1,7 @@
 parameters:
+    testsuites:
+        phpstan:
+            tasks: [phpstan]
     git_dir: .
     bin_dir: ./vendor/bin
     tasks:
@@ -14,7 +17,9 @@ parameters:
         phpunit: ~
         phplint: ~
         phpstan:
+            level: 4
             configuration: "phpstan.neon"
             ignore_patterns:
                 - "spec/"
                 - "vendor/"
+                - "test"
--- a/grumphp.yml
+++ b/grumphp.yml
@@ -1,7 +1,4 @@
 parameters:
-    testsuites:
-        phpstan:
-            tasks: [phpstan]
     git_dir: .
     bin_dir: ./vendor/bin
     tasks:
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -2,5 +2,6 @@ parameters:
     ignoreErrors:
         # Soapclient reflection is quite broken
         - '#Call to an undefined static method SoapClient::__construct().#'
-        # This is something we hope to fix in later PHP versions
+        # This is something we hope to fix in PHP 7.4
+        # https://wiki.php.net/rfc/covariant-returns-and-contravariant-parameters
         - '#Call to an undefined method Phpro\\.*(Context|RuleSet)Interface.*$#'
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,3 +1,5 @@
 parameters:
-  ignoreErrors:
-    - '#Call to an undefined static method SoapClient::__construct().#'
+    ignoreErrors:
+        - '#Call to an undefined static method SoapClient::__construct().#'
+        - '#Call to an undefined method Phpro\\.*Interface.*$#'
+        - '#Invalid type Http\\Client\\Exception to throw.#'
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,5 +1,6 @@
 parameters:
     ignoreErrors:
+        # Soapclient reflection is quite broken
         - '#Call to an undefined static method SoapClient::__construct().#'
-        - '#Call to an undefined method Phpro\\.*Interface.*$#'
-        - '#Invalid type Http\\Client\\Exception to throw.#'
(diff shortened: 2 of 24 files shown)


phpstan.neon
```diff
@@ -1,3 +1,5 @@
 parameters:
-  ignoreErrors:
-    - '#Call to an undefined static method SoapClient::__construct().#'
\ No newline at end of file
+    ignoreErrors:
+        - '#Call to an undefined static method SoapClient::__construct().#'
+        - '#Call to an undefined method Phpro\\.*Interface.*$#'
+        - '#Invalid type Http\\Client\\Exception to throw.#'
```
veewee: Actually, I think this might be a bug if you are using an async client:
http://docs.php-http.org/en/latest/httplug/exceptions.html

Maybe better to validate if its throwable and if not wrap it with a built-in httplug exception or just return the exception interface instead as suggested in the docs? To investigate how that behaves in this codebase though...

grumphp.yml
```diff
@@ -1,4 +1,7 @@
 parameters:
+    testsuites:
```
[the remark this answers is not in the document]
janvernieuwe: Ah i was still on an older version

phpstan.neon
```diff
@@ -1,3 +1,5 @@
 parameters:
-  ignoreErrors:
-    - '#Call to an undefined static method SoapClient::__construct().#'
\ No newline at end of file
+    ignoreErrors:
+        - '#Call to an undefined static method SoapClient::__construct().#'
+        - '#Call to an undefined method Phpro\\.*Interface.*$#'
```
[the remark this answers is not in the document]
janvernieuwe: RuleSetInterface has the same problem possibly.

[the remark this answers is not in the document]
veewee: Isn't it better to add 2 specific regexes with introducing comments if that's the case?

phpstan.neon
janvernieuwe: They all seem to extend their RuntimeException, wich on its turn extends \Exception, could perhaps check for that one.

janvernieuwe: Just added a throwable check for now

grumphp.yml
```diff
@@ -14,7 +14,9 @@ parameters:
         phpunit: ~
         phplint: ~
         phpstan:
+            level: 4
```
veewee: Shouldnt we add this to the neon config file as well so that you can also run the executable directly?

phpstan.neon
```diff
@@ -1,3 +1,6 @@
 parameters:
-  ignoreErrors:
-    - '#Call to an undefined static method SoapClient::__construct().#'
\ No newline at end of file
+    ignoreErrors:
+        # Soapclient reflection is quite broken
+        - '#Call to an undefined static method SoapClient::__construct().#'
+        # This is something we hope to fix in later PHP versions
```
veewee: maybe mention PHP 7.4 and link to https://wiki.php.net/rfc/covariant-returns-and-contravariant-parameters

src/Phpro/SoapClient/Middleware/Middleware.php
```diff
@@ -44,6 +45,10 @@ public function afterResponse(ResponseInterface $response): ResponseInterface
 
     public function onError(Exception $exception)
     {
+        if (!$exception instanceof Throwable) {
+            return;
```
veewee: Ignoring the exception doesnt seem acceptable to me. Maybe best to create a MiddlewareException which is throwable and which can contain an httplug exception? 
In that new exception, we can add logic to check if it's an actual exception or dislpay a standard error string?
This logic change is also something which should be covered with a unit test.

janvernieuwe: Problem is, that interface has no methods, so even getting a message from it will trigger PHPStan.
In reality all those interface implementations will be throwable for as far as i could see.

veewee: Something like this shouldn't throw an error in phpstan I guess.
````php
class MiddlewareException extends \Http\Client\Exception\TransferException
{
    public static function fromHttPlugException(\Http\Client\Exception $exception): self
    {
        $message = $exception instanceof \Throwable ? $exception->getMessage() : 'Something went wrong:'.get_class($exception);
        $previous = $exception instanceof \Throwable ? $exception : null;
        
        return new self(
            $message,
            0,
            $previous
        );
    }
}

````

grumphp.yml
janvernieuwe: Level is not supported in there it seems, there is an issue in the phpstan lib to add it, but don't think it will be in there any time soon.

PR #253, opened 2019-07-29
call cleanUpTemporaryState() if SoapFault is thrown in __soapCall fixes #252 

### LINKED ISSUE #252: ExtSoapDecoder fails to cleanUpTemporaryState on SoapFault Exception
### Bug Report

|    Q        |   A
|------------ | ------
| Version     | 1.0.3

#### Summary

\Phpro\SoapClient\Soap\Driver\ExtSoap\ExtSoapDecoder fails to call cleanUpTemporaryState() if a SoapFault Exception in __soapCall is thrown. The next SOAPcall will not contact the server but return the stored (faulty) response.

#### How to reproduce

Produce a SOAPFault and re-send the request in the same thread. You need to restart thread to clear the cache
#### Expected behavior

The Exception should be cought, and the temporary state should be cleared.

lukastaferner:


janvernieuwe:
Hi @lukastaferner , Thanks for the PR!

This looks like a good use case for `finally` to me.

lukastaferner:
Thanks for the hin @janvernieuwe, I pushed the changes!

Issue #249, opened 2019-06-14
Symfony 4.3 deprecations and soap-client


abenamer:
|    Q        |   A
|------------ | -----
| Version     | 1.0.4

### Support Question

I'm seeing the following error message: 
[info] User Deprecated: The "Phpro\SoapClient\Event\ResponseEvent" class extends "Symfony\Component\EventDispatcher\Event" that is deprecated since Symfony 4.3, use "Symfony\Contracts\EventDispatcher\Event" instead.

Looks like Symfony 4.x and soap-client aren't exactly happy with each other. Any suggestions?

veewee:
We are currently supporting Symfony 2.8 to 4.x.
This package will still keep on working untill Symfony 5.0 removes the Event class.
At that point, we should be bumping to a minimum version Symfony 4.3.

Now it's to early to bump to that version, since it is only out for a month.
Unless you have an alternative suggestion I suggest you just ignore the deprecation warning in your logs.

crashev:
Any update here?

Reverze:
When a fixes can be expected?

veewee:
We would surely accept a well-thought-out PR for this.

PR #262, opened 2019-11-22
remove decrecaptions for Symfony 4

Livda:
Remove the deprecations for Symfony 4 applications, but still maintain backward compatibility with older versions.
Resolve https://github.com/phpro/soap-client/issues/249

### DIFF
--- a/src/Phpro/SoapClient/Client.php
+++ b/src/Phpro/SoapClient/Client.php
@@ -13,7 +13,6 @@ use Phpro\SoapClient\Type\ResultProviderInterface;
 use Phpro\SoapClient\Util\XmlFormatter;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\EventDispatcher\Event as SymfonyEvent;
-use Symfony\Contracts\EventDispatcher\EventDispatcherInterface as SymfonyContractEventDispatcherInterface;
 
 /**
  * Class Client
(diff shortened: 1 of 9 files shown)


src/Phpro/SoapClient/Client.php
```diff
@@ -56,6 +57,22 @@ public function debugLastSoapRequest(): array
         ];
     }
 
+    /**
+     * For backward compatibility with Symfony 4
+     *
+     * @param string            $eventName
+     * @param SymfonyEvent|null $event
+     */
+    private function dispatch(string $eventName, SymfonyEvent $event = null): void
```
veewee: This method might break BC in webservices that provide a 'dispatch' soap method (which is quite plausible)
You could move the logic to a separate class instead and run it from the call method.

Livda: It's a private function so I don't see how it could break the BC. No child or other class can call this method. Could you provide an example where I have break something ?

Livda:
I think it's a better idea too, to use #263. It was just not there when I write my PR

src/Phpro/SoapClient/Client.php
veewee: When a soap service has a method 'dispatch', it will be added to the generated client classes.
This way you get this BC break:
https://3v4l.org/rKHWR

Livda: It won't because it's a private method. Here is a better way to see it : https://3v4l.org/ddOGk

veewee: Indeed. Thanks for pointing it out

Livda:
I found the PR #263 a bit messy, with a lot of non necessary code. So I've bump up the dependencies of my code, and updated the code to make it works but by doing so, I've break many tests and I don't know how to fix them

veewee:
Thanks for your work @Livda. 
It took me some time to come back to this PR, but I fixed the errors you had and changed the structure of the fix a bit. Feel free to take a look at
https://github.com/phpro/soap-client/pull/266

PR #263, opened 2019-11-23
Upgrade compatible version with Symfony 5.0

BPavol:
Upgrade compatible version with Symfony 5.0, create EventDispatcherAdapter for backwards compatibility

#SymfonyHackday

### DIFF
--- a/src/Phpro/SoapClient/Adapter/EventDispatcherAdapter.php
+++ b/src/Phpro/SoapClient/Adapter/EventDispatcherAdapter.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace Phpro\SoapClient\Adapter;
+
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+
+class EventDispatcherAdapter implements \Symfony\Contracts\EventDispatcher\EventDispatcherInterface
+{
+    private $dispatcher;
+
+    public function __construct(EventDispatcherInterface $dispatcher)
+    {
+        $this->dispatcher = $dispatcher;
+    }
+
+    /**
+     * Dispatches an event to all registered listeners.
+     *
+     * For BC with Symfony 4, the $eventName argument is not declared explicitly on the
+     * signature of the method. Implementations that are not bound by this BC contraint
+     * MUST declare it explicitly, as allowed by PHP.
+     *
+     * @param object $event The event to pass to the event handlers/listeners
+     * @param string|null $eventName The name of the event to dispatch. If not supplied,
+     *                               the class of $event should be used instead.
+     *
+     * @return object The passed $event MUST be returned
+     */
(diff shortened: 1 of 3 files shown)


src/Phpro/SoapClient/Adapter/EventDispatcherAdapter.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+namespace Phpro\SoapClient\Adapter;
```
veewee: Maybe better to place it in `Phpro\SoapClient\Event\Adapter` to make sure it's not about a soap adapter.

src/Phpro/SoapClient/Client.php
```diff
@@ -30,10 +31,15 @@ class Client implements ClientInterface
...
 
-    public function __construct(EngineInterface $engine, EventDispatcherInterface $dispatcher)
+    public function __construct(EngineInterface $engine, $dispatcher)
     {
         $this->engine = $engine;
         $this->dispatcher = $dispatcher;
+
+        // Backward compatibility
+        if ($dispatcher instanceof \Symfony\Component\EventDispatcher\EventDispatcherInterface) {
+            $this->dispatcher = new EventDispatcherAdapter($dispatcher);
```
veewee: This means that the dispatcher can have 2 different types. Since client is THE extension point of this package, changing the type of the dispatcher is a BC break.
I'dd suggest to move this logic to the adapter instead and wrap the dispatcher in the call method instead.

src/Phpro/SoapClient/Client.php
```diff
@@ -30,10 +31,15 @@ class Client implements ClientInterface
      */
     protected $dispatcher;
 
-    public function __construct(EngineInterface $engine, EventDispatcherInterface $dispatcher)
+    public function __construct(EngineInterface $engine, $dispatcher)
```
veewee: Is there no type that we can use? I assume Symfony 5 removes the old interface but 2 and 3 doesn't have the contract available?

src/Phpro/SoapClient/Adapter/EventDispatcherAdapter.php
```diff
@@ -0,0 +1,33 @@
...
+     *
+     * @param object $event The event to pass to the event handlers/listeners
+     * @param string|null $eventName The name of the event to dispatch. If not supplied,
+     *                               the class of $event should be used instead.
+     *
+     * @return object The passed $event MUST be returned
+     */
+    public function dispatch($event, string $eventName = null)
+    {
+        return $this->dispatcher->dispatch($eventName, $event);
```
veewee: wouldnt this break in case eventName = null? Shouldnt it fall back in `get_class($event)` or something similar instead?

Issue #272, opened 2020-01-20
__getLastRequest and __getLastResponse are empty


abryb:
![phppro_bug_last_response](https://user-images.githubusercontent.com/25817915/72725696-ae0d8700-3b86-11ea-9bab-16789f48f645.png)

PR #273, opened 2020-01-20
fix issue-#272 - empty __getLastRequest and __getLastResponse

abryb:


veewee:
Thanks for logging and fixing the issue!

PR #285, opened 2020-03-03
Update Psr7Converter.php

### LINKED ISSUE #284: Response body not rewound in Psr7Converter
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no? (first-time user :))
| Version     | v1.1.2

#### Summary

See 

https://github.com/phpro/soap-client/blob/390acc42bb70e7d7bf7cc543beb81524fa322ab4/src/Phpro/SoapClient/Soap/HttpBinding/Converter/Psr7Converter.php#L64-L66

This procudes an empty string for us. When changing to `(string) $response->getBody()` all is fine.

This is related to the internal pointer which is eof at this point, so it needs to be rewound. Hence we use `__toString` instead.

See also similar issues like https://github.com/OpenAPITools/openapi-generator/issues/4541 which i fixed the same way: https://github.com/OpenAPITools/openapi-generator/pull/4744

ro0NL:
fixes #284 :)

### DIFF
--- a/src/Phpro/SoapClient/Soap/HttpBinding/Converter/Psr7Converter.php
+++ b/src/Phpro/SoapClient/Soap/HttpBinding/Converter/Psr7Converter.php
@@ -62,7 +62,7 @@ class Psr7Converter
     public function convertSoapResponse(ResponseInterface $response): SoapResponse
     {
         return new SoapResponse(
-            $response->getBody()->getContents()
+            (string) $response->getBody()
         );
     }
 }


ro0NL:
im not aware where the resonse body is consumed actually, but it does seem to happen by default (no custom middlewares) :confused: maybe im the first one to notice?

https://github.com/phpro/soap-client/blob/390acc42bb70e7d7bf7cc543beb81524fa322ab4/src/Phpro/SoapClient/Soap/Handler/HttPlugHandle.php#L85-L87

adding `dd($psr7Response->getBody()->getContents(), (string) $psr7Response->getBody());` below `$psrResponse` gives us

```
HttPlugHandle.php on line 86:
""

HttPlugHandle.php on line 86:
"<s:Envelope xmlns:s="http://schemas.xmlsoap.org/soap/envelope/"><s:Body xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/ ▶"
```

veewee:
@ro0NL : it might depend on the client and the version you are using. It is possible that the http client isnt rewinding the response stream. Maybe that's the case here?

Anyways, the provided fix should be ok since the PSR spec requires a rewing before rendering as string.

ro0NL:
@veewee  i figured it out.


> no custom middlewares

i lied :sweat_smile:  we have custom guzzle middleware for logging http traffic (bound to the HandlerStack). This message format includes the response body string :)

janvernieuwe:
Sorry it took so long.
Going to release asap.

PR #287, opened 2020-03-07
RemoveEmptyNodesMiddleware: reduce till evelope body

### LINKED ISSUE #286: RemoveEmptyNodesMiddleware removes root node
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 1.1.2

#### Summary

`RemoveEmptyNodesMiddleware` is great, however when sending an "unfilled object" (e.g. all properties are null), the node removal bubbles up to the root node.

This breaks the soap format in general, and produces an application error during `Client::call()`:

```
Warning: DOMDocument::loadXML(): Start tag expected, '&lt;' not found in Entity, line: 2
```

I think we should always send an "empty soap envelope" , thus preserving `<SOAP-ENV:Envelope /`> and perhaps an empty body/method node as well?

Input xml:

```
<?xml version="1.0" encoding="UTF-8"?>
<SOAP-ENV:Envelope xmlns:SOAP-ENV="http://schemas.xmlsoap.org/soap/envelope/" xmlns:ns1="http://method" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
   <SOAP-ENV:Body>
      <ns1:method>
         <arg xsi:nil="true" />
      </ns1:method>
   </SOAP-ENV:Body>
</SOAP-ENV:Envelope>
```

Output xml:

```
<?xml version="1.0" encoding="UTF-8"?>
```

Expected output XML:

```
<?xml version="1.0" encoding="UTF-8"?>
<SOAP-ENV:Envelope xmlns:SOAP-ENV="http://schemas.xmlsoap.org/soap/envelope/" xmlns:ns1="http://method" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" />
```

or

```
<?xml version="1.0" encoding="UTF-8"?>
<SOAP-ENV:Envelope xmlns:SOAP-ENV="http://schemas.xmlsoap.org/soap/envelope/" xmlns:ns1="http://method" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
   <SOAP-ENV:Body />
</SOAP-ENV:Envelope>
```

or

```
<?xml version="1.0" encoding="UTF-8"?>
<SOAP-ENV:Envelope xmlns:SOAP-ENV="http://schemas.xmlsoap.org/soap/envelope/" xmlns:ns1="http://method" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
   <SOAP-ENV:Body>
      <ns1:method />
   </SOAP-ENV:Body>
</SOAP-ENV:Envelope>
```

Thoughts? This should improve debugging as well :)

ro0NL:
fixes #286

### DIFF
--- a/src/Phpro/SoapClient/Middleware/RemoveEmptyNodesMiddleware.php
+++ b/src/Phpro/SoapClient/Middleware/RemoveEmptyNodesMiddleware.php
@@ -32,7 +32,7 @@ class RemoveEmptyNodesMiddleware extends Middleware
 
     private function getNotNodes(Xml $xml): ?\DOMNodeList
     {
-        $notNodes = $xml->xpath('//*[not(node())]');
+        $notNodes = $xml->xpath('//soap:Envelope/*//*[not(node())]');
         if (!$notNodes->length) {
             return null;
         }


janvernieuwe:
Can you provide a test to verify that this works as expected and to prevent regressions in the future?

janvernieuwe:
Thanks for the contribution!

PR #290, opened 2020-04-07
Rewind the body before decoding using the conventioned tostring on body

janvernieuwe:
When using unit tests / logging  and not rewinding the body stream before decoding gives null results.


janvernieuwe:
See #285

Issue #267, opened 2019-12-06
PHP 74 support


veewee:
Unit tests fail on PHP74. To investigate and fix ....
See Travis:
https://travis-ci.org/phpro/soap-client/jobs/621513849?utm_medium=notification&utm_source=github_status

veewee:
Blocked by

    php-vcr/php-vcr#289
    php-vcr/php-vcr#293

janvernieuwe:
Going to close this as this is resolved sufficiently for now.
We only need a tag from php-vcr, but its not blocking anymore

PR #293, opened 2020-04-15
Update HTTPlug dependencies + only enable php-vcr where it makes sense

### LINKED ISSUE #275: Upgrade Http-plug
We are using an older version of HTTPlug at the moment.
This issue is for upgrading those packages.

> php-http/httplug required in this library is 1.1 while the library itself is at 2.x now

veewee:
* Fixes the error in #267, but still doesnt work on PHP 7.4 since php-vcr is not on-track.
* Describes how to use httplug v2 in the docs. (Fixes #275)


composer.json
```diff
@@ -26,14 +26,14 @@
...
     "php-http/discovery": "^1.7",
-    "php-http/guzzle6-adapter": "^1.1.1",
-    "php-http/httplug": "^1.1",
+    "php-http/guzzle6-adapter": "^2.0.1",
+    "php-http/httplug": "^2.1",
     "php-http/message": "^1.8",
     "php-http/message-factory": "^1.0",
     "php-http/mock-client": "^1.3",
-    "php-vcr/php-vcr": "~1.4.4",
+    "php-vcr/php-vcr": "dev-issues/289 as 1.4.5",
```
veewee, line 36: Don't realy like this solution, but it works until a stable version is tagged..

driehle:
I would love to see this PR merged an released as Httplug v2 would allow using the [RequestSeekableBodyPlugin](https://github.com/php-http/client-common/blob/master/src/Plugin/RequestSeekableBodyPlugin.php) and the [ResponseSeekableBodyPlugin](https://github.com/php-http/client-common/blob/master/src/Plugin/ResponseSeekableBodyPlugin.php). Currently, calling `$client->debugLastSoapRequest()` leads to the following error when using php-http/socket-client:

```
Fatal error: Uncaught Http\Client\Socket\Exception\StreamException: 
This stream is not seekable in [...]\vendor\php-http\socket-client\src\Stream.php:134
```

driehle:
Well, fyi, I have merged #293, #295 and #296 together [here](https://github.com/driehle/soap-client/tree/hotfix/20200422) and luckily Travis has [nothing to complain](https://travis-ci.org/github/driehle/soap-client/builds/678331261). :-)

veewee:
FYI: you can already use httplug 2 without any change in this package. Only the documentation changed. 
We'll review, merge and release soon!

janvernieuwe:
Imho, we should get rid of php-vcr, since it it not maintained at all anymore.
That we have to use a PR as the main version says enough.

composer.json
janvernieuwe: I've been waiting on a merge for months, as but we could do this as a stop-gap measure, I only hope that @morozov won't delete hes PR then.

janvernieuwe:
https://twitter.com/5andshark/status/1253662217494462466?s=20 :crossed_fingers:

driehle:
@janvernieuwe How about you clone the php-vcr repository in the phpro namespace, especially the dev-issues/289 branch, and refer it from there in your composer.json? That would make the whole thing a bit more stable from the perspective of soap-client users, until php-vcr finally releases a new version. Even if they don't that would still get you some more time in terms of getting rid of php-vcr alltogether.

veewee:
There is no other vcr for ext-soap so  it would force people to use httplug in which there is a decent replacement based on middlewares.
The vcr library is not blocking for this package, it is just being used in the testsuite atm.

janvernieuwe:
Now we only need a release from php-vcr, and we're good.
I'll switch to master branch commit for now after an update to this branch.

drupol:
Why can't we just user the PSR18 for HTTP Client ? I think HTTPlug is now deprecated in favor of it, no ?

janvernieuwe:
> Why can't we just user the PSR18 for HTTP Client ? I think HTTPlug is now deprecated in favor of it, no ?

We can do that in another PR, first I want to get this merged and release a new version today.

drupol:
Fair enough :-)

composer.json
morozov: I see you switched to `dev-master`, so I'm going to delete the branch.

PR #297, opened 2020-04-25
add `void` reserved keywords

Pchol:


drupol:
:+1:

janvernieuwe:
Thanks for the contribution! I know we don't have a test for this class yet.
It would be a very nice addition.

PR #301, opened 2020-06-09
fixed required version of psr/http-message in httplug docs

freezy-sk:
psr/http-message does not exists in 1.8 version


veewee:
Thanks!

PR #302, opened 2020-06-12
Duplicate types enhencements

### LINKED ISSUE #300: Error while generating the Types
Hi,

I'm using this library to build a client for an internal service.

However, I'm having an issue while building the types.

See here the different relevant config files and log files.

**Soap Client config file**: https://del.dog/ifingaivai.txt
**Log**: https://del.dog/mitevaluno.txt
**Console command**: `./vendor/bin/soap-client generate:types -o --config=config/soap-client-Adonis.php -vvv`

I succeeded to find which class is causing the issue and at the end of the processing, I noticed a singularity with the `File` type, 2 files are generated for it: `File.php` and `File.php.backup`

`File.php`: https://del.dog/ogruxalofi.txt
`File.php.backup`: https://del.dog/apamaristu.txt

It looks like the correct file is `File.php.backup` and that there are some missing stuff in `File.php`.

I don't know where to start to debug the issue, any hint/clue is very welcome.

|    Q        |   A
|------------ | -----
| Version     | 1.2.0

veewee:
Fixes #300

I made something that allows you to manipulate the metadata before types are generated.
To try it out, I also added 2 possible strategies for fixing duplicate types.

- IntersectDuplicateTypesStrategy (default) -> this merges all duplicate types into 1 big type
- RemoveDuplicateTypesStrategy

This seems to do the trick.
Currently it matches on exact case match. This might not work if you have similar types in different cases. For example a "file" and a "File" type. 

I was thinking about other strategies as well, but there is just no unique identifier like an XSD namespace. This means I could for example add an increment number, but then the code will break once items are moved inside the XSD (which is suboptimal). Another option was to make a hash of the properties, but that would also break if the type changes. Therefore, I think it is imposible to generate separate classes per type with ext-soap.

:question: **Unless anyone has any better idea?**

TODO:

- [x] Tests
- [x] Documentation
- [x] Improved duplicates detector (case insensitive, special chars, ...)

### DIFF
--- a/docs/drivers/ext-soap.md
+++ b/docs/drivers/ext-soap.md
@@ -50,7 +50,7 @@ $typemap->add(new MyTypeConverter());
 ### Duplicate types
 
 Ext-soap does not add any namespace or unique identifier to the types it knows.
-You can read more about this in the [known ext-soap issues](../known-issues/ext-soap.md) section.
+You can read more about this in the [known ext-soap issues](../known-issues/ext-soap.md#duplicate-typenames) section.
 Therefore, we added some strategies to deal with duplicate types:
 
 **IntersectDuplicateTypesStrategy**
@@ -62,6 +62,7 @@ This duplicate types strategy will merge all duplicate types into one big type w
 **RemoveDuplicateTypesStrategy**
 
 This duplicate types strategy will remove all duplicate types it finds.
+
 
 
 You can overwrite the strategy on the `ExtSoapOptions` object:
--- a/docs/drivers/ext-soap.md
+++ b/docs/drivers/ext-soap.md
@@ -76,7 +76,7 @@ use Phpro\SoapClient\Soap\Engine\Metadata\MetadataOptions;
 
 $options = ExtSoapOptions::defaults($wsdl)
     ->withMetaOptions(function (MetadataOptions $options): MetadataOptions {
-        return $options->withMethodsManipulator(
+        return $options->withTypesManipulator(
             new RemoveDuplicateTypesStrategy()
         );
     });
--- a/docs/drivers/ext-soap.md
+++ b/docs/drivers/ext-soap.md
@@ -75,7 +75,7 @@ use Phpro\SoapClient\Soap\Driver\ExtSoap\Metadata\Manipulators\DuplicateTypes\Re
 use Phpro\SoapClient\Soap\Engine\Metadata\MetadataOptions;
 
 $options = ExtSoapOptions::defaults($wsdl)
-    ->withMetaOptions(function (MetadataOptions $options): MetadataOptions {
+    ->withMetadataOptions(function (MetadataOptions $options): MetadataOptions {
         return $options->withTypesManipulator(
             new RemoveDuplicateTypesStrategy()
         );
--- a/docs/drivers/ext-soap.md
+++ b/docs/drivers/ext-soap.md
@@ -44,3 +44,39 @@ $options = ExtSoapOptions::defaults($wsdl, ['location' => 'http://somedifferents
 $typemap = $options->getTypeMap();
 $typemap->add(new MyTypeConverter());
 ```
+
+## Dealing with ext-soap issues
+
+### Duplicate types
+
+Ext-soap does not add any namespace or unique identifier to the types it knows.
+You can read more about this in the [known ext-soap issues](../known-issues/ext-soap.md) section.
+Therefore, we added some strategies to deal with duplicate types:
+
+**IntersectDuplicateTypesStrategy**
+
+Enabled by default when using `ExtSoapOptions::defaults()`.
+
+This duplicate types strategy will merge all duplicate types into one big type which contains all properties.
+
+**RemoveDuplicateTypesStrategy**
+
+This duplicate types strategy will remove all duplicate types it finds.
+
+
+You can overwrite the strategy on the `ExtSoapOptions` object:
+
+```php
+<?php
+
+use Phpro\SoapClient\Soap\Driver\ExtSoap\ExtSoapOptions;
+use Phpro\SoapClient\Soap\Driver\ExtSoap\Metadata\Manipulators\DuplicateTypes\RemoveDuplicateTypesStrategy;
+use Phpro\SoapClient\Soap\Engine\Metadata\MetadataOptions;
+
+$options = ExtSoapOptions::defaults($wsdl)
+    ->withMetaOptions(function (MetadataOptions $options): MetadataOptions {
+        return $options->withMethodsManipulator(
+            new RemoveDuplicateTypesStrategy()
+        );
+    });
+```
--- a/docs/drivers/metadata.md
+++ b/docs/drivers/metadata.md
@@ -12,7 +12,7 @@ On top of that, we provide some generic tools that can be useful.
 
 ## LazyInMemoryMetadata
 
-This class will make sure that the metadata is only collected once and stores it in memory for a second access.
+This class will make sure that the metadata are only collected once and stores it in memory for a second access.
 
 Example output:
 
--- a/docs/drivers/metadata.md
+++ b/docs/drivers/metadata.md
@@ -3,7 +3,7 @@
 The metadata part of the driver knows what objects and functions are inside the soap service.
 This can be done by parsing the WSDL file.
 Every driver has its own way of collecting this information.
-On top of that, we provide some generic tools that can be usefull.
+On top of that, we provide some generic tools that can be useful.
 
 # Built-in metadata tools
 
--- a/docs/drivers/metadata.md
+++ b/docs/drivers/metadata.md
@@ -1,0 +1,55 @@
+# Driver metadata
+
+The metadata part of the driver knows what objects and functions are inside the soap service.
+This can be done by parsing the WSDL file.
+Every driver has its own way of collecting this information.
+On top of that, we provide some generic tools that can be usefull.
+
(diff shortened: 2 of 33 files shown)


src/Phpro/SoapClient/Soap/Driver/ExtSoap/Metadata/Manipulators/DuplicateTypes/IntersectDuplicateTypesStrategy.php
```diff
@@ -0,0 +1,47 @@
...
+use Phpro\SoapClient\Soap\Engine\Metadata\Collection\PropertyCollection;
+use Phpro\SoapClient\Soap\Engine\Metadata\Collection\TypeCollection;
+use Phpro\SoapClient\Soap\Engine\Metadata\Manipulators\TypesManipulatorInterface;
+use Phpro\SoapClient\Soap\Engine\Metadata\Model\Type;
+
+final class IntersectDuplicateTypesStrategy implements TypesManipulatorInterface
+{
+    public function __invoke(TypeCollection $allTypes): TypeCollection
+    {
+        return new TypeCollection(...array_values($allTypes->reduce(
```
janvernieuwe, line 16: Perhaps this could be refactored a little for improved readability?

veewee: Hello @janvernieuwe, to me it is quite readable and clear (especially once I added tests). Feel free to tell me what you think is not that readable or how I could improve.

janvernieuwe: It would just be minor, like more the function to a class method, split things up a bit.
It can be left as is, it's just a suggestion.

docs/drivers/ext-soap.md
```diff
@@ -44,3 +44,39 @@ $options = ExtSoapOptions::defaults($wsdl, ['location' => 'http://somedifferents
...
+<?php
+
+use Phpro\SoapClient\Soap\Driver\ExtSoap\ExtSoapOptions;
+use Phpro\SoapClient\Soap\Driver\ExtSoap\Metadata\Manipulators\DuplicateTypes\RemoveDuplicateTypesStrategy;
+use Phpro\SoapClient\Soap\Engine\Metadata\MetadataOptions;
+
+$options = ExtSoapOptions::defaults($wsdl)
+    ->withMetaOptions(function (MetadataOptions $options): MetadataOptions {
+        return $options->withMethodsManipulator(
+            new RemoveDuplicateTypesStrategy()
```
drupol, line 80: Be careful, `RemoveDuplicateTypesStrategy` is not an instance of `MethodsManipulatorInterface`.

docs/drivers/ext-soap.md
```diff
@@ -44,3 +44,39 @@ $options = ExtSoapOptions::defaults($wsdl, ['location' => 'http://somedifferents
...
+```php
+<?php
+
+use Phpro\SoapClient\Soap\Driver\ExtSoap\ExtSoapOptions;
+use Phpro\SoapClient\Soap\Driver\ExtSoap\Metadata\Manipulators\DuplicateTypes\RemoveDuplicateTypesStrategy;
+use Phpro\SoapClient\Soap\Engine\Metadata\MetadataOptions;
+
+$options = ExtSoapOptions::defaults($wsdl)
+    ->withMetaOptions(function (MetadataOptions $options): MetadataOptions {
+        return $options->withMethodsManipulator(
```
drupol, line 79:
```suggestion
        return $options->withTypesManipulator(
```

veewee: Thanks, overlooked that one!

docs/drivers/metadata.md
```diff
@@ -0,0 +1,55 @@
+# Driver metadata
+
+The metadata part of the driver knows what objects and functions are inside the soap service.
+This can be done by parsing the WSDL file.
+Every driver has its own way of collecting this information.
+On top of that, we provide some generic tools that can be usefull.
```
drupol, line 6:
```suggestion
On top of that, we provide some generic tools that can be useful.
```

docs/drivers/metadata.md
```diff
@@ -0,0 +1,55 @@
...
+
+# Built-in metadata tools
+
+- [LazyInMemoryMetadata](#lazyinmemorymetadata)
+- [ManipulatedMetadata](#manipulatedmetadata)
+
+## LazyInMemoryMetadata
+
+This class will make sure that the metadata is only collected once and stores it in memory for a second access.
```
drupol, line 15:
```suggestion
This class will make sure that the metadata are only collected once and stores it in memory for a second access.
```

drupol:
This is great, massive improvements, can't wait to see this merged :-)

src/Phpro/SoapClient/Soap/Driver/ExtSoap/Metadata/Detector/DuplicateTypeNamesDetector.php
```diff
@@ -0,0 +1,33 @@
...
+     * @param TypeCollection $types
+     *
+     * @return string[]
+     */
+    public function __invoke(TypeCollection $types): array
+    {
+        return array_keys(
+            array_filter(
+                array_count_values($types->map(
+                    static function (Type $type) {
```
drupol, line 23:
```suggestion
                    static function (Type $type): string {
```

src/Phpro/SoapClient/Soap/Driver/ExtSoap/Metadata/Manipulators/DuplicateTypes/IntersectDuplicateTypesStrategy.php
```diff
@@ -0,0 +1,48 @@
...
+            function (array $result, Type $type) use ($allTypes) {
```
drupol, line 18:
```suggestion
            function (array $result, Type $type) use ($allTypes): array {
```

src/Phpro/SoapClient/Soap/Engine/Metadata/MetadataOptions.php
```diff
@@ -0,0 +1,55 @@
...
+     * @var MethodsManipulatorInterface
+     */
+    private $methodsManipulator;
+
+    /**
+     * @var TypesManipulatorInterface
+     */
+    private $typesManipulator;
+
+    public function __construct()
```
drupol, line 24: How about using DI principle here ?

Once it's done, we can get rid of the 2 `getters` as well. I always find it weird to use `getters` and `withers`.

Because of the `getters`, we could potentially taint the object properties and then, the immutability would be broken. 

```suggestion
    public function __construct(MethodManipulatorInterface $methodManipulator, TypeManipulatorInterface $typeManipulator)
```

veewee: I made both possible by adding a static constructor.
In this case : it is more a configurable object which should be as flexible as possible IMO.
The manipulators should only have an `__invoke` method, so they should never be mutable.

src/Phpro/SoapClient/Soap/Driver/ExtSoap/ExtSoapOptions.php
```diff
@@ -114,6 +133,18 @@ public function disableWsdlCache(): self
         return $this;
     }
 
+    public function withMetaOptions(callable $manipulator): self
```
drupol, line 136: Just to be consistent ?

```suggestion
    public function withMetadataOptions(callable $manipulator): self
```

src/Phpro/SoapClient/Soap/Engine/Metadata/MetadataOptions.php
drupol: Ok fair enough.

I just find it odd to wrap `MethodManipulatorInterface` and `TypeManipulatorInterface` in such object. I made a POC locally and we could get rid of `MetadataOptions` class pretty easily.

That's just my 2 cents.

src/Phpro/SoapClient/Soap/Driver/ExtSoap/ExtSoapOptions.php
veewee: Fixed

src/Phpro/SoapClient/Soap/Engine/Metadata/MetadataOptions.php
veewee: Agree, it's a bit of a strange class... We could get rid of it by making the MetadataInterface explicit in the engine's factory classes? Or what did you have in mind?

drupol: I would just get rid of it and use `MethodManipulatorInterface` and `TypeManipulatorInterface` at the place.

I don't think there is a need to wrap them in a class `MetadataOptions`, unless you plan to add more things to it later on, but I doubt.

veewee: I played around with it a bit more and decided to leave it as-is for now.
The problem is that I want to specify the "defaults" somewhere, but don't want to change too many places at the same time.
This intermediate object does exactly that. Although I agree that at some point we might kick it out for direct injection of a metadata object. Time will tell ... :)

PR #307, opened 2020-07-01
 Fix a problem if the class has not yet been generated

gitlog:


veewee:
Hello @gitlog,

It looks like the issue is in the `TypeChecker`. Instead of removing the line, it is a better solution to implement the method that checks for known classes:

https://github.com/phpro/soap-client/blob/master/src/Phpro/SoapClient/CodeGenerator/Util/TypeChecker.php#L46-L50

Let me know if you are willing to give it a shot.

PR #310, opened 2020-07-06
Feature: Add ClosureAssembler

gitlog:
`ClosureAssembler` - This implementation makes it possible use closure/callback

Usage:

```php
   $config ->addRule(new Rules\AssembleRule(new Assembler\ClosureAssembler(
        function(ContextInterface $context): bool {
            return $context instanceof TypeContext;
        },
        function(ContextInterface $context) {

            // some code...
        })
    ))
```

### DIFF
--- a/src/Phpro/SoapClient/CodeGenerator/Assembler/ClosureAssembler.php
+++ b/src/Phpro/SoapClient/CodeGenerator/Assembler/ClosureAssembler.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace Phpro\SoapClient\CodeGenerator\Assembler;
+
+use Phpro\SoapClient\CodeGenerator\Context\ContextInterface;
+use Phpro\SoapClient\CodeGenerator\Context\TypeContext;
+use Phpro\SoapClient\Exception\AssemblerException;
+
+/**
+ * Class FinalClassAssembler
+ *
+ * @package Phpro\SoapClient\CodeGenerator\Assembler
+ */
+class ClosureAssembler implements AssemblerInterface
+{
+
+    /**
+     * @var callable
+     */
+    private $canAssembleClosure;
+
+    /**
+     * @var callable
+     */
+    private $assembleClosure;
+
+
+    /**
+     * ClosureAssembler constructor.
+     * @param callable $canAssembleClosure
+     * @param callable $assembleClosure
+     */
+    public function __construct(callable $canAssembleClosure, callable $assembleClosure)
+    {
+        $this->canAssembleClosure = $canAssembleClosure;
+        $this->assembleClosure = $assembleClosure;
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


src/Phpro/SoapClient/CodeGenerator/Assembler/ClosureAssembler.php
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /**
+     * @param ContextInterface $context
+     *
+     * @return bool
+     */
+    public function canAssemble(ContextInterface $context): bool
+    {
+        return call_user_func($this->canAssembleClosure, $context);
```
veewee, line 46:
```suggestion
        return ($this->canAssembleClosure)($context);
```

src/Phpro/SoapClient/CodeGenerator/Assembler/ClosureAssembler.php
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /**
+     * @param ContextInterface|TypeContext $context
+     */
+    public function assemble(ContextInterface $context)
+    {
+        call_user_func($this->assembleClosure, $context);
```
veewee, line 54:
```suggestion
        ($this->assembleClosure)($context);
```

src/Phpro/SoapClient/CodeGenerator/Assembler/ClosureAssembler.php
```diff
@@ -0,0 +1,56 @@
...
+
+
+    /**
+     * ClosureAssembler constructor.
+     * @param callable $canAssembleClosure
+     * @param callable $assembleClosure
+     */
+    public function __construct(callable $canAssembleClosure, callable $assembleClosure)
+    {
+        $this->canAssembleClosure = $canAssembleClosure;
```
veewee, line 35: Instead of having 2 callables. Would it make sense to always return `true` in the `canAssemble` method?
That way you only need to pass one callable and you can return early based on context.

Would that make sense?

src/Phpro/SoapClient/CodeGenerator/Assembler/ClosureAssembler.php
```diff
@@ -0,0 +1,56 @@
+<?php
+
+namespace Phpro\SoapClient\CodeGenerator\Assembler;
+
+use Phpro\SoapClient\CodeGenerator\Context\ContextInterface;
+use Phpro\SoapClient\CodeGenerator\Context\TypeContext;
+use Phpro\SoapClient\Exception\AssemblerException;
+
+/**
+ * Class FinalClassAssembler
```
veewee, line 10: You can drop these class docblocks IMO

veewee:
Closing because of inactivity. Feel free to finish and reopen.

PR #312, opened 2020-07-08
Feature: Add ArrayAssembler

gitlog:
Implementation of IteratorAggregate (https://www.php.net/manual/ru/class.iteratoraggregate.php), ArrayAccess (https://www.php.net/manual/ru/class.arrayaccess.php), Countable (https://www.php.net/manual/ru/class.countable.php) interfaces to provide accessing objects as arrays.
When we has generated class with IteratorAssembler - we can to add only one element to this array. To add more elements, must use ArrayAccess interface. With Countable interface will be available element's counting, with IteratorAggregate - iterating array elements in loop.

### DIFF
--- a/src/Phpro/SoapClient/CodeGenerator/Assembler/ArrayAssembler.php
+++ b/src/Phpro/SoapClient/CodeGenerator/Assembler/ArrayAssembler.php
@@ -1,0 +1,235 @@
+<?php
+
+namespace Phpro\SoapClient\CodeGenerator\Assembler;
+
+use Phpro\SoapClient\CodeGenerator\Context\ContextInterface;
+use Phpro\SoapClient\CodeGenerator\Context\TypeContext;
+use Phpro\SoapClient\CodeGenerator\Model\Property;
+use Phpro\SoapClient\CodeGenerator\LaminasCodeFactory\DocBlockGeneratorFactory;
+use Phpro\SoapClient\Exception\AssemblerException;
+use Laminas\Code\Generator\ClassGenerator;
+use Laminas\Code\Generator\MethodGenerator;
+
+/**
+ * Class IteratorAssembler
+ *
+ * @package Phpro\SoapClient\CodeGenerator\Assembler
+ */
+class ArrayAssembler implements AssemblerInterface
+{
+    /**
+     * @param ContextInterface $context
+     *
+     * @return bool
+     */
+    public function canAssemble(ContextInterface $context): bool
+    {
+        return $context instanceof TypeContext;
+    }
+
+    /**
+     * @param ContextInterface|TypeContext $context
+     *
+     * @throws AssemblerException
+     */
+    public function assemble(ContextInterface $context)
+    {
+        $class = $context->getClass();
+        $properties = $context->getType()->getProperties();
+        $firstProperty = count($properties) ? current($properties) : null;
+
+        try {
+
+            $interfaceAssembler = new InterfaceAssembler(\IteratorAggregate::class);
+            if ($interfaceAssembler->canAssemble($context)) {
+                $interfaceAssembler->assemble($context);
+            }
+
+            $interfaceAssembler = new InterfaceAssembler(\ArrayAccess::class);
+            if ($interfaceAssembler->canAssemble($context)) {
+                $interfaceAssembler->assemble($context);
+            }
+
+            $interfaceAssembler = new InterfaceAssembler(\Countable::class);
+            if ($interfaceAssembler->canAssemble($context)) {
+                $interfaceAssembler->assemble($context);
+            }
+
+            if ($firstProperty) {
+
+            	$this->implementConstruct($class, $firstProperty);
+                $this->implementOffsetExists($class, $firstProperty);
+                $this->implementOffsetGet($class, $firstProperty);
+                $this->implementOffsetSet($class, $firstProperty);
+                $this->implementOffsetUnset($class, $firstProperty);
+                $this->implementCount($class, $firstProperty);
+            }
+
+        } catch (\Exception $e) {
+            throw AssemblerException::fromException($e);
+        }
+    }
+
+    /**
+     * @param ClassGenerator $class
+     * @param Property       $firstProperty
+     *
+     * @throws \Laminas\Code\Generator\Exception\InvalidArgumentException
+     */
+    private function implementConstruct(ClassGenerator $class, Property $firstProperty)
+    {
+        $methodName = '__construct';
+        $class->removeMethod($methodName);
+        $class->addMethodFromGenerator(MethodGenerator::fromArray([
+            'name' => $methodName,
+            'visibility' => MethodGenerator::VISIBILITY_PUBLIC,
+    		'body' => sprintf('$this->%s = [];', $firstProperty->getName()),
+			'docblock' => DocBlockGeneratorFactory::fromArray([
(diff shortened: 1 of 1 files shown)


src/Phpro/SoapClient/CodeGenerator/Assembler/ArrayAssembler.php
```diff
@@ -0,0 +1,235 @@
...
+    /**
+     * @param ContextInterface|TypeContext $context
+     *
+     * @throws AssemblerException
+     */
+    public function assemble(ContextInterface $context)
+    {
+        $class = $context->getClass();
+        $properties = $context->getType()->getProperties();
+        $firstProperty = count($properties) ? current($properties) : null;
```
veewee, line 39: It is better to skip if the first property does not exist.
Otherwise the class implements the interfaces, but not the methods.

src/Phpro/SoapClient/CodeGenerator/Assembler/ArrayAssembler.php
```diff
@@ -0,0 +1,235 @@
...
+
+        try {
+
+            $interfaceAssembler = new InterfaceAssembler(\IteratorAggregate::class);
+            if ($interfaceAssembler->canAssemble($context)) {
```
veewee, line 44: It does not contain the `getIterator()` method.
I'dd suggest to drop that interface hete and use the `IteratorAssembler` directly instead.

src/Phpro/SoapClient/CodeGenerator/Assembler/ArrayAssembler.php
```diff
@@ -0,0 +1,235 @@
...
+        }
+    }
+
+    /**
+     * @param ClassGenerator $class
+     * @param Property       $firstProperty
+     *
+     * @throws \Laminas\Code\Generator\Exception\InvalidArgumentException
+     */
+    private function implementConstruct(ClassGenerator $class, Property $firstProperty)
```
veewee, line 79: It is better to set the properties default to `[]` instead of using the constructor.
Now you will cause an issue if you mix it with the constructor assembler.
This method can be removed IMO

src/Phpro/SoapClient/CodeGenerator/Assembler/ArrayAssembler.php
```diff
@@ -0,0 +1,235 @@
...
+            'visibility' => MethodGenerator::VISIBILITY_PUBLIC,
+            'docblock' => DocBlockGeneratorFactory::fromArray([
+                'tags' => [
+                    [
+                        'name' => 'param',
+                        'description' => 'mixed $offset An offset to check for',
+                    ],
+                    [
+                        'name' => 'return',
+                        'description' => 'boolean true on success or false on failure',
```
veewee, line 117: you could add the strict type here as well

src/Phpro/SoapClient/CodeGenerator/Assembler/ArrayAssembler.php
```diff
@@ -0,0 +1,235 @@
...
+     */
+    private function implementOffsetGet(ClassGenerator $class, Property $firstProperty)
+    {
+        $methodName = 'offsetGet';
+        $class->removeMethod($methodName);
+
+        $class->addMethodFromGenerator(MethodGenerator::fromArray([
+            'name' => $methodName,
+            'parameters' => ['name' => 'offset'],
+            'body' => sprintf('return $this->%s[$offset];', $firstProperty->getName()),
```
veewee, line 139: Maybe safer to check if the offset exists first.

src/Phpro/SoapClient/CodeGenerator/Assembler/ArrayAssembler.php
```diff
@@ -0,0 +1,235 @@
...
+     */
+    private function implementOffsetSet(ClassGenerator $class, Property $firstProperty)
+    {
+        $methodName = 'offsetSet';
+        $class->removeMethod($methodName);
+
+        $class->addMethodFromGenerator(MethodGenerator::fromArray([
+            'name' => $methodName,
+            'parameters' => [['name' => 'offset'], ['name' => 'value']],
+            'body' => sprintf("if(isset(\$this->%s[\$offset])) {\n\t\$this->%s[] = \$value;\n} else {\n\t\$this->%s[\$offset] = \$value; \n}", $firstProperty->getName(), $firstProperty->getName(), $firstProperty->getName()),
```
veewee, line 170: Why did you add the if statement? You can just set the property index to the value right?

src/Phpro/SoapClient/CodeGenerator/Assembler/ArrayAssembler.php
```diff
@@ -0,0 +1,235 @@
...
+     */
+    private function implementOffsetUnset(ClassGenerator $class, Property $firstProperty)
+    {
+        $methodName = 'offsetUnset';
+        $class->removeMethod($methodName);
+
+        $class->addMethodFromGenerator(MethodGenerator::fromArray([
+            'name' => $methodName,
+            'parameters' => ['name' => 'offset'],
+            'body' => sprintf('unisset($this->%s[$offset]);', $firstProperty->getName()),
```
veewee, line 197: There is no `unisset` method in php

src/Phpro/SoapClient/CodeGenerator/Assembler/ArrayAssembler.php
```diff
@@ -0,0 +1,235 @@
...
+
+        $class->addMethodFromGenerator(MethodGenerator::fromArray([
+            'name' => $methodName,
+            'body' => sprintf('return count($this->%s);', $firstProperty->getName()),
+            'visibility' => MethodGenerator::VISIBILITY_PUBLIC,
+            'docblock' => DocBlockGeneratorFactory::fromArray([
+                'tags' => [
+                    [
+                        'name' => 'return',
+                        'description' => sprintf("%s Return count of elements", $firstProperty->getType()),
```
veewee, line 229: Count doesnt return the type, it returns an int.
you can also use this as real return type.

src/Phpro/SoapClient/CodeGenerator/Assembler/ArrayAssembler.php
gitlog: How to set the properties default to [] ?

src/Phpro/SoapClient/CodeGenerator/Assembler/ArrayAssembler.php
gitlog: Fix in next commit 

Out:

```php
    /**
     * @param mixed $offset An offset to check for
     */
    public function offsetSet($offset, $value)
    {
        if(!isset($offset)) {
        	$this->Prop[] = $value;
        } else {
        	$this->Prop[$offset] = $value; 
        }
    }
```
Use:

```php

$obj[] = $val1;

$obj[10] = $val2;

```

src/Phpro/SoapClient/CodeGenerator/Assembler/ArrayAssembler.php
gitlog: Fix in next commit

veewee:
Closing because of inactivity. Feel free to finish and reopen.

veewee:
Closing this one. In v3, the type system will be much better.
Normalizing the type doesn't seem something we want to be doing from this package.

PR #314, opened 2020-08-28
Fixes binary response error in tests

janvernieuwe:
You could get this as body in the tests:
image

This fixes that, so that valid xml is returned instead.

### DIFF
--- a/src/Phpro/SoapClient/Soap/HttpBinding/Converter/Psr7ToLastRequestInfoConverter.php
+++ b/src/Phpro/SoapClient/Soap/HttpBinding/Converter/Psr7ToLastRequestInfoConverter.php
@@ -36,7 +36,7 @@ class Psr7ToLastRequestInfoConverter
             );
 
             $requestHeaders = trim($requestParts[0] ?? '');
-            $requestBody = $requestParts[1] ?? '';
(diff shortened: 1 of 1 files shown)


src/Phpro/SoapClient/Soap/HttpBinding/Converter/Psr7ToLastRequestInfoConverter.php
```diff
@@ -36,7 +36,7 @@ public function convert(
             );
 
             $requestHeaders = trim($requestParts[0] ?? '');
```
veewee, line 38: Is this still the best way to fetch the formatter headers?

veewee (approved):

src/Phpro/SoapClient/Soap/HttpBinding/Converter/Psr7ToLastRequestInfoConverter.php
janvernieuwe: I looked into it, the getHeaders returns it as an array of `['header-name' => ['value1', 'value2']]`
That might not be optimal, and the current solution seems to be working fine for now.

PR #322, opened 2020-11-06
Do not set type hint and return type "mixed" on PHP < 8

user890104:
Mixed is not a valid type on PHP < 8, it was introduced in 8.0.0 by https://wiki.php.net/rfc/mixed_type_v2
When useReturnType (getter) and/or useTypeHints (setter) are enabled, the assembler sets "mixed" both in docblocks and as PHP types in the code. The latter causes issues on PHP < 8.

### DIFF
--- a/src/Phpro/SoapClient/CodeGenerator/Model/Property.php
+++ b/src/Phpro/SoapClient/CodeGenerator/Model/Property.php
@@ -71,6 +71,20 @@ class Property
     }
 
     /**
+     * @return string|null
+     */
+    public function getCodeReturnType(): ?string
+    {
+        $type = $this->getType();
+
+        if ($type === 'mixed' && PHP_VERSION_ID < 80000) {
+            return null;
+        }
+
+        return $type;
+    }
+
+    /**
      * @return string
      */
     public function getterName(): string
--- a/src/Phpro/SoapClient/CodeGenerator/Assembler/GetterAssembler.php
+++ b/src/Phpro/SoapClient/CodeGenerator/Assembler/GetterAssembler.php
@@ -59,7 +59,7 @@ class GetterAssembler implements AssemblerInterface
                     'parameters' => [],
                     'visibility' => MethodGenerator::VISIBILITY_PUBLIC,
                     'body'       => sprintf('return $this->%s;', $property->getName()),
-                    'returntype' => $this->options->useReturnType() ? $property->getType() : null,
+                    'returntype' => $this->options->useReturnType() ? $property->getCodeReturnType() : null,
                     'docblock'   => DocBlockGeneratorFactory::fromArray([
                         'tags' => [
                             [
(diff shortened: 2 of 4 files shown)


janvernieuwe:
Hi @user890104 ,
Thank you for your PR!

After reviewing it, I personally think it would be better to split the return type methods into 2, one for the doc blocks, and one for the code return type like this:

```php
    public function getDocblockReturnType(): string
    {
        if (Normalizer::isKnownType($this->type)) {
            return $this->type;
        }

        return '\\'.$this->namespace.'\\'.Normalizer::normalizeClassname($this->type);
    }

    public function getCodeReturnType(): ?string
    {
        $type = $this->getDocblockReturnType();

        if ($type === 'mixed' && PHP_VERSION_ID <= 80000) {
            return null;
        }

        return $type;
    }
```

This way the behavior is easy to test, no duplicate checks, and easy if we need to exclude another type.

user890104:
Hi @janvernieuwe ,
I totally agree with that, I will push an updated solution later.

src/Phpro/SoapClient/CodeGenerator/Model/Property.php
```diff
@@ -70,6 +70,20 @@ public function getType(): string
...
     }
 
+    /**
+     * @return string|null
+     */
+    public function getCodeReturnType(): ?string
+    {
+        $type = $this->getType();
+
+        if ($type === 'mixed' && PHP_VERSION_ID <= 80000) {
```
janvernieuwe, line 80: I made an error here, should be < instead of <=

janvernieuwe:
Here is a test I was working on btw:

```php
<?php

namespace PhproTest\SoapClient\Unit\CodeGenerator\Model;

use Phpro\SoapClient\CodeGenerator\Model\Property;
use PHPUnit\Framework\TestCase;

/**
 * Class UseAssemblerTest
 *
 * @package PhproTest\SoapClient\Unit\CodeGenerator\Model
 */
class PropertyTest extends TestCase
{
    /**
     * @test
     */
    public function it_returns_null_type_pre_php8(): void
    {
        if (PHP_VERSION_ID >= 80000) {
            self::markTestSkipped('Pre PHP 8 only');
        }
        $property = new Property('test', 'mixed', 'App');
        self::assertNull($property->getCodeReturnType());
        self::assertEquals('mixed', $property->getType());
    }

    /**
     * @test
     */
    public function it_returns_mixed_type_post_php8(): void
    {
        if (PHP_VERSION_ID < 80000) {
            self::markTestSkipped('Post PHP 8 only');
        }
        $property = new Property('test', 'mixed', 'App');
        self::assertEquals('mixed', $property->getCodeReturnType());
        self::assertEquals('mixed', $property->getType());
    }
}

```

src/Phpro/SoapClient/CodeGenerator/Model/Property.php
user890104: Fixed

user890104:
I've added your test to the latest commit, and seems to work fine.

janvernieuwe (approved):

janvernieuwe:
Thanks!

Issue #323, opened 2020-11-13
"No class is set" when generating types or client (laminas-code 3.5.0 BC break)


edudobay:
Hello! When generating the types and the client class for a SOAP service (with the command `soap-client generate:types` or `soap-client generate:client`), I got error messages and the code generation didn't go through.

When generating types, I would get repeated instances of the error message `No class is set` and no type file was generated. The client generation failed in the same way with the same message, additionally referencing the line where the error was thrown (`FileGenerator.php:350`). With the help of a debugger I traced the problem to these lines up the call stack:

https://github.com/phpro/soap-client/blob/df5d3af128e0758269cc773b11a5d5e1757016d1/src/Phpro/SoapClient/CodeGenerator/TypeGenerator.php#L42

https://github.com/phpro/soap-client/blob/df5d3af128e0758269cc773b11a5d5e1757016d1/src/Phpro/SoapClient/CodeGenerator/ClientGenerator.php#L44

I came up with a dirty workaround and managed to get my types and client generated:

```diff
+        try {
            $class = $file->getClass() ?: new ClassGenerator();
+        } catch (ClassNotFoundException $exception) {
+            $class = new ClassGenerator();
+        }
```

Here are the [WSDL](https://github.com/phpro/soap-client/files/5534067/wsdl.txt) and [soap-client.php](https://github.com/phpro/soap-client/files/5534068/soap-client.php.txt) that I used for reproducing this issue. Note: this is a random WSDL I found on the internet (https://tomd.xyz/wsdl/). Originally I had this issue with another WSDL (that I'm not sure would be ok for publicly sharing).

Also, this is the output for `composer show`: [composer_show.txt](https://github.com/phpro/soap-client/files/5534084/composer_show.txt) (should be mostly latest stable releases, as this was a fresh project).

edudobay:
Update: this change in laminas-code changes the behavior for `FileGenerator::getClass()`, introducing that exception: https://github.com/laminas/laminas-code/commit/cc2adf24d3dac4bca644c5f872c40c01840929e9. This is the corresponding PR: https://github.com/laminas/laminas-code/pull/43

Downgrading to laminas-code 3.4.1 fixes the issue. So this seems to be a BC break introduced by laminas-code 3.5.0.

veewee:
@janvernieuwe : doh! :)


The fix you provided looks like the right way to go. Would you care to create a PR @edudobay?

janvernieuwe:
[...]

PR #324, opened 2020-11-13
Fix BC break from laminas-code 3.5.0 (#323)

edudobay:
Fixes https://github.com/phpro/soap-client/issues/323.

I'm unsure about the best way to write the tests for this change. In the first commit, I just duplicated code. In the second one, I tried to remove some duplication, but couldn't find much to do while keeping tests clear and readable — I think that extracting _all_ duplication would make tests brittle and obscure. I'd appreciate some ideas on this :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         "guzzlehttp/guzzle": "^6.4.1",
         "guzzlehttp/promises": "^1.3.1",
         "guzzlehttp/psr7": "^1.6.1",
-        "laminas/laminas-code": "^3.4.0",
+        "laminas/laminas-code": "^3.5.0",
         "nyholm/psr7": "^1.3",
         "php-http/client-common": "^2.1",
         "php-http/discovery": "^1.7",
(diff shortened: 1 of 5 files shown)


edudobay:
The test suite fails on all builds with laminas-code 3.4.* — that is, all those with `DEPENDENCIES=low` and also all PHP 7.2 builds, because laminas-code 3.5.* only supports PHP >= 7.3.

laminas-code 3.4.* does not declare `ClassNotFoundException` and this yields two errors:
1. PHPStan complains that a caught class does not exist
1. an instance of that class must be created to assert correct behavior when that exception is thrown (this is a test case I added)

I can't yet figure out a solution for the PHPStan error. As for the second one, I thought of some ideas and none of them seemed good enough to me. I would appreciate suggestions on which direction to follow:
* remove the test case I added
* try to rewrite it so that is becomes "higher-level" and unaware of these implementation details and work in either version
* annotate it so that it is only run in PHP >= 7.3; but that does not account for `DEPENDENCIES=low`
* conditionally create that class, if it does not exist

PR #325, opened 2020-11-16
Use fromString instead of fromStream when fetching the request body

dhzavann:
Trying to use WsseMiddleware with Guzzle 6 does not seem to work without this change; $stream->getContents() in ::fromStream() returns an empty string.


janvernieuwe (approved):
Looks good,
Will merge in the next session.

Thank you for your contribution

PR #328, opened 2020-11-27
Add php80 support

veewee:


veewee:
Marked as 1.4.0: 
there are some small bc changes like added final keywords and dropped support for PHP72 and symfony 34.

I dont really consider them a BC break since we did not change any code.

grumphp.yml
```diff
@@ -9,7 +9,7 @@ grumphp:
         phpunit: ~
         phplint: ~
         phpstan:
-            level: 3
+            level: 1
```
janvernieuwe, line 12: image

janvernieuwe (approved):
Too bad we have to start on level 1 again with phpstan, can't be helped I guess.
Good to get rid of a ton of old laminas checks! :+1:

PR #341, opened 2021-01-26
define composer conflict with php-http/guzzle6-adapter

staabm:
because of "BC breaking" kind-of changes in https://github.com/phpro/soap-client/commit/e8f2b55761bc52272632ac39af2517fc7c8116a0

see https://github.com/phpro/soap-client/commit/e8f2b55761bc52272632ac39af2517fc7c8116a0#commitcomment-46282702


janvernieuwe:
Thanks!

Issue #220, opened 2019-02-06
Generated Response Classes are not used in Client


just-tom:
|    Q        |   A
|------------ | -----
| Version     | 0.7.11

Hi guys, Im using the generator on some open WSDL's to test its functionality before integrating it with the service I need to. It's working great so far, except one problem. The response given is always a MixedResult, so the generate call methods throw an error expecting an instance of a generated Response class. Im not too sure where to go to solve the issue. 

The WSDL file im testing with is:

> http://www.xignite.com/xcurrencies.asmx?WSDL

An example call is:
```
/**
     * @param RequestInterface|Type\ListActiveCurrencies $parameters
     * @return ResultInterface|Type\ListActiveCurrenciesResponse
     * @throws SoapException
     */
    public function listActiveCurrencies(\App\Type\ListActiveCurrencies $parameters) : \App\Type\ListActiveCurrenciesResponse
    {
        return $this->call('ListActiveCurrencies', $parameters);
    }
```
Error Thrown is:

> PHP Fatal error:  Uncaught TypeError: Return value of App\CurrencyClient::listActiveCurrencies() must be an instance of App\Type\ListActiveCurrenciesResponse, instance of Phpro\SoapClient\Type\MixedResult returned in /Users/thomasburman/temp/soap-scaff/src/CurrencyClient.php:30
> Stack trace:
> #0 /Users/thomasburman/temp/soap-scaff/run.php(39): App\CurrencyClient->listActiveCurrencies(Object(App\Type\ListActiveCurrencies))
> #1 /Users/thomasburman/temp/soap-scaff/run.php(48): MyVatClient->getCompanyVatDetails('GB', '282194785')
> #2 {main}

Generated config is:
```
<?php

use Phpro\SoapClient\CodeGenerator\Assembler;
use Phpro\SoapClient\CodeGenerator\Rules;
use Phpro\SoapClient\CodeGenerator\Config\Config;

return Config::create()
    ->setWsdl('http://www.xignite.com/xcurrencies.asmx?WSDL')
    ->setTypeDestination('src/Type')
    ->setTypeNamespace('App\Type')
    ->setClientDestination('src')
    ->setClientName('CurrencyClient')
    ->setClientNamespace('App')
    ->setClassMapDestination('src')
    ->setClassMapName('CurrencyClassmap')
    ->setClassMapNamespace('App')
[...]

PR #342, opened 2021-02-02
clarify HttPlugHandle example

RafaelKr:
clarify example to prevent MixedResult error like in https://github.com/phpro/soap-client/issues/220


RafaelKr:
@veewee could you please have a quick look at this? This could save other devs some hours :)

veewee:
Thanks! Looks good

PR #351, opened 2021-04-21
Added a new WsaMiddleware2005 Middleware to support WSA w3c 2005

staabm:
without the fix we had errors like 

> PHP SOAP Header does not match HTTP SOAP Action

since for soap 1.2 it seems it is required to use WSANS_2005

### DIFF
--- a/docs/middlewares.md
+++ b/docs/middlewares.md
@@ -53,7 +53,7 @@ Internally it used the [wse-php package of robrichards](https://github.com/robri
 which is a well known library that is used by many developers.
 The middleware is a light wrapper that makes it easy to use in your application.
 
-In case you need WSA w3c 2005 based Web Service Addressing, you should use WsaMiddleware2005
+In case you need [WSA w3c 2005 based Web Service Addressing](https://www.w3.org/TR/2005/CR-ws-addr-soap-20050817/#soaphttp), you should use WsaMiddleware2005.
 
 **Dependencies**
 ```sh
--- a/docs/middlewares.md
+++ b/docs/middlewares.md
@@ -13,7 +13,7 @@ Next, you can use one of the built-in middlewares:
 
 - [BasicAuthMiddleware](#basicauthmiddleware)
 - [NtlmMiddleware](#ntlmmiddleware) (deprecated)
-- [WsaMiddleware](#wsamiddleware)
+- [WsaMiddleware / WsaMiddleware2005](#wsamiddleware)
 - [WsseMiddleware](#wssemiddleware)
 - [RemoveEmptyNodesMiddleware](#removeemptynodesmiddleware)
 - [WS-I Compliance/QuotedSoapActionMiddleware](#ws-i-compliancequotedsoapactionmiddleware)
@@ -53,6 +53,8 @@ Internally it used the [wse-php package of robrichards](https://github.com/robri
 which is a well known library that is used by many developers.
 The middleware is a light wrapper that makes it easy to use in your application.
 
+In case you need WSA w3c 2005 based Web Service Addressing, you should use WsaMiddleware2005
+
 **Dependencies**
 ```sh
 composer require robrichards/wse-php:^2.0
@@ -60,7 +62,9 @@ composer require robrichards/wse-php:^2.0
 
 **Usage**
 ```php
-$handler->addMiddleware(new WsaMiddleware());
+$handler->addMiddleware(new WsaMiddleware()); // WSA xmlsoap 2004 based Web Service Addressing
+
+$handler->addMiddleware(new WsaMiddleware2005()); // for WSA w3c 2005 based Web Service Addressing
 ```
 
 
--- a/src/Phpro/SoapClient/Middleware/WsaMiddleware.php
+++ b/src/Phpro/SoapClient/Middleware/WsaMiddleware.php
@@ -11,6 +11,7 @@ use RobRichards\WsePhp\WSASoap;
 class WsaMiddleware extends Middleware
 {
     const WSA_ADDRESS_ANONYMOUS = 'http://schemas.xmlsoap.org/ws/2004/08/addressing/role/anonymous';
+    const WSA_ADDRESS2005_ANONYMOUS = 'http://www.w3.org/2005/08/addressing';
 
     private $address;
 
@@ -28,7 +29,11 @@ class WsaMiddleware extends Middleware
     {
         $xml = SoapXml::fromStream($request->getBody());
 
-        $wsa = new WSASoap($xml->getXmlDocument());
+        if ($this->address === self::WSA_ADDRESS2005_ANONYMOUS) {
+            $wsa = new WSASoap($xml->getXmlDocument(), WSASoap::WSANS_2005);
+        } else {
+            $wsa = new WSASoap($xml->getXmlDocument());
+        }
         $wsa->addAction(SoapActionDetector::detectFromRequest($request));
         $wsa->addTo((string) $request->getUri());
         $wsa->addMessageID();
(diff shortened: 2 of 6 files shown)


src/Phpro/SoapClient/Middleware/WsaMiddleware.php
```diff
@@ -11,6 +11,7 @@
 class WsaMiddleware extends Middleware
 {
     const WSA_ADDRESS_ANONYMOUS = 'http://schemas.xmlsoap.org/ws/2004/08/addressing/role/anonymous';
+    const WSA_ADDRESS2005_ANONYMOUS = 'http://www.w3.org/2005/08/addressing';
```
staabm, line 14: maybe this could/should be renamed to 

```suggestion
    const WSA_ADDRESS2005 = 'http://www.w3.org/2005/08/addressing';
```

since the rule/anonymous part is no longer part of the url

veewee: You are confusing 2 different settings in here:

There is the WSA namespace, which could be either:

* WSA xmlsoap 2004 : 'http://schemas.xmlsoap.org/ws/2004/08/addressing'
* WSA w3c 2005 : 'http://www.w3.org/2005/08/addressing'

This namespace is used to prefix the added xml elements according to the selected version.
For example:

`xmlns:wsa="http://schemas.xmlsoap.org/ws/2004/08/addressing"`

Meaning that all `<wsa:x />` elements point to the 2004 xmlsoap version.


Besides the namespace, there is the anonymous Reply-To field.
This is to specify where the webserver needs to send its reply to.
If it is anonymous, it takes current session to send the response to.

This endpoint is different for wsa xmlsoap 2004 and wsa w3c 2005:

* WSA xmlsoap 2004 : 'http://schemas.xmlsoap.org/ws/2004/08/addressing/role/anonymous'
* WSA w3c 2005 : 'http://www.w3.org/2005/08/addressing/anonymous'

More info, see : https://www.w3.org/TR/2005/CR-ws-addr-soap-20050817/#soaphttp

src/Phpro/SoapClient/Middleware/WsaMiddleware.php
```diff
@@ -28,7 +29,11 @@ public function beforeRequest(callable $handler, RequestInterface $request): Pro
     {
         $xml = SoapXml::fromStream($request->getBody());
 
-        $wsa = new WSASoap($xml->getXmlDocument());
+        if ($this->address === self::WSA_ADDRESS2005_ANONYMOUS) {
```
veewee, line 32: Since the reply address can also be something different than the anonymous endpoint,
this check is not a valid solution.

I am wondering: wouldn't it make more sense to provide a separate WSA2005 middleware instead?
It would make it a lot easier IMO.

veewee (commented):
Thanks for your PR!
I've added some remarks. Can you take a look at them?

We'll also require unit tests in order to get this feature merged in, to make sure it works as expected.

src/Phpro/SoapClient/Middleware/WsaMiddleware.php
clxmstaab: thx for your input.

I have separated the 2005-middleware into its own file was a separate test.
hopefully thats what you had in mind and I got you right.

test/PhproTest/SoapClient/Unit/Middleware/WsaMiddlewareTest2005.php
```diff
@@ -0,0 +1,118 @@
...
+
+        // Check defaults:
+        $this->assertEquals('myaction', $xml->xpath('//soap:Header/wsa:Action')->item(0)->nodeValue);
+        $this->assertEquals('/endpoint', $xml->xpath('//soap:Header/wsa:To')->item(0)->nodeValue);
+        $this->assertMatchesRegularExpression(
+            '/^uuid:[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i',
+            $xml->xpath('//soap:Header/wsa:MessageID')->item(0)->nodeValue
+        );
+        $this->assertEquals(
+            WsaMiddleware::WSA_ADDRESS_ANONYMOUS,
```
veewee, line 98: Shouldn't this be the WSA 2005 anonymous endpoint?

veewee:
In general, it looks good! Can you also add documentation similar to the regular wsa middleware?

https://github.com/phpro/soap-client/blob/master/docs/middlewares.md#wsamiddleware

docs/middlewares.md
```diff
@@ -53,14 +53,18 @@ Internally it used the [wse-php package of robrichards](https://github.com/robri
 which is a well known library that is used by many developers.
 The middleware is a light wrapper that makes it easy to use in your application.
 
+In case you need WSA w3c 2005 based Web Service Addressing, you should use WsaMiddleware2005
```
clxmstaab, line 56:
```suggestion
In case you need [WSA w3c 2005 based Web Service Addressing](https://www.w3.org/TR/2005/CR-ws-addr-soap-20050817/#soaphttp), you should use WsaMiddleware2005.
```

test/PhproTest/SoapClient/Unit/Middleware/WsaMiddlewareTest2005.php
clxmstaab: hmm makes me feel the new test is not run. not sure I am right.

fixed the typo.

clxmstaab: had a typo in the testclass-name. fixed.

test/PhproTest/SoapClient/Unit/Middleware/WsaMiddlewareTest2005.php
```diff
@@ -0,0 +1,118 @@
...
+     *
+     * @return SoapXml
+     */
+    private function fetchSoapXml($soapBody): SoapXml
+    {
+        $xml = new \DOMDocument();
+        $xml->loadXML($soapBody);
+
+        $soapXml = new SoapXml($xml);
+        $soapXml->registerNamespace('wsa', 'http://schemas.xmlsoap.org/ws/2005/08/addressing');
```
clxmstaab, line 114: should this be changed to `WsaMiddleware2005::WSA_ADDRESS2005_ANONYMOUS` (and a anlog change in `WsaMiddlewareTest`)?

veewee: You could use`WSASoap::WSANS_2005` for that.
(Not the anonymous one provided in the middleware, but the xmlns one from WSASoap.)

Not that it will ever change though.

veewee (approved):

test/PhproTest/SoapClient/Unit/Middleware/WsaMiddlewareTest2005.php
clxmstaab: ahh good point, fixed.

staabm:
btw: I still have the feeling that the newly added test is not run on CI.

I ran it manually locally with ` php vendor/bin/phpunit test/PhproTest/SoapClient/Unit/Middleware/WsaMiddlewareTest2005.php -vvv` and everthing went green.

staabm:
Ping @veewee

veewee:
My week was challenging.
I'm gonna run some tests myself to make sure it works soon.

Thanks!

veewee:
Seems to work just fine (tests inclusive!) Thanks!

clxmstaab:
> Seems to work just fine (tests inclusive!) Thanks!

we use it already in one of our systems for a few days, I guess its fine :).

Thanks for your feedback! great project <3

PR #352, opened 2021-04-28
Add option to remove docblocks from getters and setters

simivar:
In this PR we add `withDocBlocks` option to Getters and Setters generators the same way we have it in the Constructor generator

### DIFF
--- a/docs/code-generation/assemblers.md
+++ b/docs/code-generation/assemblers.md
@@ -119,11 +119,11 @@ Example output:
     }
 ```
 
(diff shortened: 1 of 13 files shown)


veewee:
Thanks for the PR!
Looks nice at first sight, but asking @janvernieuwe to take a second look.

janvernieuwe (approved):
Thanks for this PR, it looks good!
Maybe in the future we can ask in the wizard to select docblocks or not.

PR #354, opened 2021-04-28
Upgrade Laminas Code to ^4.2

simivar:


negoziator (approved):
Nice! Also with the formatting 👍🏼

veewee:
Thanks for the PR.

It does not seem to work completely:

```
Error: Call to an undefined static method Laminas\Code\Generator\FileGenerator::fromReflectedFileName().
Error: Call to an undefined static method Laminas\Code\Generator\FileGenerator::fromReflectedFileName().
Error: Call to an undefined static method Laminas\Code\Generator\FileGenerator::fromReflectedFileName().
```

Can you take a look at these issues?


Would it make sense to bind it to 4.0 instead of 4.2 or are there specific things added in 4.2?

This PR implies that we have to drop support for PHP 73, which I totally agree with! :)

simivar:
@veewee do we actually need `patchExistingFile()` methods of `GenerateClassmapCommand`, `GenerateClientCommand` and `GenerateTypesCommand` classes? From what I see no matter what modifications are done to these classes they will be overwritten by whatever was generated by `generate()` method. Am I missing something or those could be removed altogether?

veewee:
The patch logic loads the PHP file as it is. Next the code generation will run.
Meaning that if a property got removed, it will still be available in that class.

I personally never use that patch logic, since it is doing strange things from time to time.
Maybe time to drop it and to always overwrite classes instead?

What do you think @janvernieuwe  ?

janvernieuwe:
If people still want to maintain the previously generated code, they could still do that using version control (undo those deletes).
I don't use the patch functionality personally either, i just delete everything and regenerate.

So yes, I vote to remove it since it is quirky and doesn't provide much added value.

PR #356, opened 2021-04-29
Change properties to be protected instead of private

negoziator:
This will make it possible to extend the `WsseMiddleware` with a custom middleware class.

This is already possible - but that will require you to extend the base `Middleware` class.

With this you can choose to only override a single method of need.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 vendor
 composer.lock
 .phpunit.result.cache
+.idea
--- a/src/Phpro/SoapClient/Middleware/WsseMiddleware.php
+++ b/src/Phpro/SoapClient/Middleware/WsseMiddleware.php
@@ -15,62 +15,62 @@ class WsseMiddleware extends Middleware
     /**
      * @var string
      */
-    private $privateKeyFile;
+    protected $privateKeyFile;
 
     /**
      * @var string
      */
-    private $publicKeyFile;
+    protected $publicKeyFile;
 
     /**
      * @var string
      */
-    private $serverCertificateFile = '';
+    protected $serverCertificateFile = '';
 
     /**
      * @var int
      */
-    private $timestamp = 3600;
+    protected $timestamp = 3600;
 
     /**
      * @var bool
      */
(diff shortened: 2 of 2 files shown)


veewee:
Thanks for the PR @negoziator.
The WSSE middleware is quite complex as-is. I would not recommend extending it.
In fact, I rather make it final.

I do think there is room for improvement in that specific middleware (even though I got it working on 5-6 projects).
Is there anything specific you want to change on it?

negoziator:
Thanks for your reply @veewee  

Yes i agree with you! - It's pretty complex.
However - there's some good stuff going on in there - and good helper functions that solves lots of problems.

I'm sitting on a project right now - i have used 4 days, trying to customize the middleware to do some specific stuff.
- please note, that i'm not very good at encryption.

The thing is. We are sending requests to the danish government - and they are very very strict on the order of items in the document.

Also ; for instance they want `ds:Signature` to be inside the `wsse:Security` ( we implemented that in our extended class )

the only method that needs to be extended is the `beforeRequest`.

That will make it possible to move things around etc.

note: i still don't have any luck in making them ( the danish government ) accepting my requests.
I'm currently looking into their latest message:

 "Make sure the x.509-certificate is inserted in the `wsse:BinarySecurityToken` as a Base64 encoded string."
Having a lot of fun here 🎉

negoziator:
In case you wanna know. 

I'm trying to make my request look like this ( This is and example of a request against their test-system )

https://github.com/skat/emcs-b2b-ws/blob/master/sample/request-test-system.xml

veewee:
Soap services like this are the reason why people hate SOAP ....
Good luck making it work!

If there is anything specific we can improve in the wsse middleware, feel free to open up another issue.

My personal vision : I would love to recreate the wsse middleware from scratch, making it work more like the interface in SoapUI:
https://www.soapui.org/docs/soapui-projects/ws-security/

This way you have more control on which element is injected where and you can make more advanced rules on which specific tags you want to apply the encryption.

Sadly, time is not in my favour ... maybe once I retire in 33 years :)

PR #363, opened 2021-04-30
Add docblocks question to Config Wizard command

simivar:
As mentioned by @janvernieuwe in #352 I've added a question to Wizard command whether methods should be generated with a docblock. It simplifies it a little as we have `withDocblocks` setting on many levels (for getters, setters, and the constructor).


janvernieuwe (approved):
Awesome!
Thanks a bunch for yet another good contribution!

PR #364, opened 2021-04-30
Remove patching from Generate commands

simivar:
As discussed in #354 we can remove "patching" existing files from Generate commands as it is quirky and doesn't provide much value and only complicates the codebase.

### DIFF
--- a/src/Phpro/SoapClient/Util/Filesystem.php
+++ b/src/Phpro/SoapClient/Util/Filesystem.php
@@ -62,30 +62,4 @@ class Filesystem
         $this->ensureDirectoryExists(\dirname($path));
         file_put_contents($path, $content);
     }
-
-    /**
-     * @param string $file
-     */
-    public function createBackup(string $file)
-    {
-        if (!$this->fileExists($file)) {
-            throw new RuntimeException('Could not create a backup from a non existing file: '.$file);
-        }
-
-        $backupFile = preg_replace('{\.backup$}', '', $file).'.backup';
-        copy($file, $backupFile);
-    }
-
-    /**
-     * @param string $file
-     */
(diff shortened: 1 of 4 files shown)


veewee:
Thanks for the pr. Will take a deeper look soon.
This will introduce a bc break. Maybe time to list some things we want to change for v2 @janvernieuwe ?

simivar:
If I think correctly you mean it's BC break because we've deleted a `protected` method that could've been extended by someone. If that's the case what do you think about, for now, removing only the body contents so we could get it to 1.x version, marking it as `@deprecated`, and then removing it altogether in 2.x?

veewee:
Yes indeed, but on top of that - it is a BC break for the people who might rely on this feature to work this specific way. After the change, the cli tool will work in a different way. This might result in broken code downstream.

veewee:
@simivar
I've pointed this PR to a new 2.0.0 branch in which we will introduce this change.
We could also accept a PR for v1.0.0 that introduces deprecation warnings. But that logging might get lost in the output of the CLI command - so for me it is also fine to describe it in an upgrading section when releasing v0.2.0

src/Phpro/SoapClient/Util/Filesystem.php
```diff
@@ -62,30 +62,4 @@ public function putFileContents(string $path, $content)
         $this->ensureDirectoryExists(\dirname($path));
         file_put_contents($path, $content);
     }
-
-    /**
-     * @param string $file
-     */
-    public function createBackup(string $file)
```
veewee, line 69: Removing the patching part is one thing. But not providing backups might break something in an irriversable way...

open question : should we still provide (or ask for creating) backup files?

/cc @janvernieuwe

Issue #379, opened 2021-06-25
better static analysis support


staabm:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

while class-generation classes implementing `IteratorAggregate` should get a few annotations, to make php static analysis tools (psalm, phpstan,..) aware of the actual types beeing wrapped.

as is the generator, produces code like

```
class ArrayOfSalesOrderLinesWS implements IteratorAggregate
{

    /**
     * @var \WebConnector\Order\Type\SalesOrderLinesWS
     */
    private $Sales_Order_Lines_WS;

    /**
     * @return \ArrayIterator
     */
    public function getIterator()
    {
        return new \ArrayIterator(is_array($this->Sales_Order_Lines_WS) ? $this->Sales_Order_Lines_WS : []);
    }
```

for static analysis tools, this construct hides the actual types beeing wrapped behind the `IteratorAggregate`.

if the code beeing generated would look like e.g.
```
/**
 * @phpstan-implements IteratorAggregate<int, SalesOrderLinesWS>
 */
class ArrayOfSalesOrderLinesWS implements IteratorAggregate
{

    /**
     * @var \WebConnector\Order\Type\SalesOrderLinesWS
     */
    private $Sales_Order_Lines_WS;

    /**
     * @return \ArrayIterator
     */
    public function getIterator()
    {
        return new \ArrayIterator(is_array($this->Sales_Order_Lines_WS) ? $this->Sales_Order_Lines_WS : []);
    }
``` 

phpstan could make use of this additional knowledge, see https://phpstan.org/r/337a116c-75c2-4ba1-abf8-3b3f04424d0d

staabm:
example of how it is right now, and how it could be:

https://phpstan.org/r/a539bc17-e946-4df0-939a-083eba57237f

clxmstaab:
after a bit of research, I guess the class which needs tweaking is https://github.com/phpro/soap-client/blob/master/src/Phpro/SoapClient/CodeGenerator/Assembler/IteratorAssembler.php

veewee:
Sounds nice. One of the issues is that we are currently not able to detect if a type is an array or a single occurence.
See: https://github.com/phpro/soap-client/blob/master/docs/known-issues/ext-soap.md#occurs

The properties are generated seperately from the iterator.
[...]

PR #380, opened 2021-06-28
Added generic types to IteratorAggregate in IteratorAssembler

staabm:
closes https://github.com/phpro/soap-client/issues/379

### DIFF
--- a/src/Phpro/SoapClient/CodeGenerator/Assembler/IteratorAssembler.php
+++ b/src/Phpro/SoapClient/CodeGenerator/Assembler/IteratorAssembler.php
@@ -76,10 +76,23 @@ class IteratorAssembler implements AssemblerInterface
                         [
                             'name' => 'return',
                             'description' => '\\ArrayIterator'
+                        ],
+                        [
+                            'name' => 'phpstan-return',
+                            'description' => '\\ArrayIterator<int, '. $firstProperty->getType() .'>'
                         ]
                     ]
                 ])
             ])
         );
+
+        $class->setDocBlock(DocBlockGeneratorFactory::fromArray([
+            'tags' => [
+                [
+                    'name' => 'phpstan-implements',
+                    'description' => '\\IteratorAggregate<int, '. $firstProperty->getType() .'>'
+                ]
+            ]
+        ]));
     }
 }
--- a/src/Phpro/SoapClient/CodeGenerator/Assembler/IteratorAssembler.php
+++ b/src/Phpro/SoapClient/CodeGenerator/Assembler/IteratorAssembler.php
@@ -79,11 +79,11 @@ class IteratorAssembler implements AssemblerInterface
                         ],
                         [
                             'name' => 'phpstan-return',
-                            'description' => '\\ArrayIterator<int, '. $firstProperty->getType() .'>'
+                            'description' => '\\ArrayIterator<array-key, '. $firstProperty->getType() .'>'
                         ],
                         [
                             'name' => 'psalm-return',
-                            'description' => '\\ArrayIterator<int, '. $firstProperty->getType() .'>'
+                            'description' => '\\ArrayIterator<array-key, '. $firstProperty->getType() .'>'
                         ]
                     ]
                 ])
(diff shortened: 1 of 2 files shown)


staabm:
> Sounds nice. One of the issues is that we are currently not able to detect if a type is an array or a single occurence.
> See: [`master`/docs/known-issues/ext-soap.md#occurs](https://github.com/phpro/soap-client/blob/master/docs/known-issues/ext-soap.md?rgh-link-date=2021-06-28T07%3A48%3A53Z#occurs)

I am not yet sure when and how this influences the changes at hand.

 
> The properties are generated seperately from the iterator.
> So it is possible to add an `@implements` attribute on the iterator, but it might not be possibly to type the property in a correct way. This will still result in static analyzer complaints.

the changes at hand now generate the following, which seems good enough for us:

grafik


> 
> I'dd also rather add a generic annotation than a psalm and/or phpstan specific one.
> Normally they both accept `@implements`.
> 

since phpstan and psalm use the annotations of the other tool if present, I don't see a problem with using the `phpstan` (or `psalm`-prefix).

we could still use the `@implements` without a prefix, but to get max compat on the `getIterator` return type, having a prefixed one would be preferable, so tools which do not support generics don't have problems with it:

grafik

veewee:
> I am not yet sure when and how this influences the changes at hand.

Check the `@var` annotation of the property. It is not typed as something iterable.
This will always result in static analyzer issues:
https://psalm.dev/r/add3f7c288

> since phpstan and psalm use the annotations of the other tool if present, I don't see a problem with using the phpstan (or psalm-prefix).

Did not know this. OK for me. Should we also add the one for psalm as well for completeness sake?

> we could still use the @implements without a prefix, but to get max compat on the getIterator return type, having a prefixed one would be preferable, so tools which do not support generics don't have problems with it.

Not sure what tools you are referring to. Everything I use knows about this syntax :-)


/ping @janvernieuwe for a second look.

staabm:
> Not sure what tools you are referring to. Everything I use knows about this syntax :-)

I can think of phpdoc generation tools or other AST based tools which work on phpdocs.

> Check the `@var` annotation of the property. It is not typed as something iterable.
> This will always result in static analyzer issues:
> [psalm.dev/r/add3f7c288](https://psalm.dev/r/add3f7c288)

isn't this problem already resolved with the IteratorAssembler 

grafik

veewee:
> isn't this problem already resolved with the IteratorAssembler ?
since the PR only changes the code generated by IteratorAssembler it should not affect code, which doesn't use it?

No, see https://psalm.dev/r/add3f7c288.

> ERROR: InvalidReturnStatement - 17:16 - The inferred type 'array<empty, empty>' does not match the declared return type 'Traversable<int, stdClass>' for X::getIterator

> ERROR: InvalidReturnType - 13:16 - The declared return type 'Traversable<int, stdClass>' for X::getIterator is incorrect, got 'array<empty, empty>'

src/Phpro/SoapClient/CodeGenerator/Assembler/IteratorAssembler.php
```diff
@@ -76,10 +76,31 @@ private function implementGetIterator(ClassGenerator $class, Property $firstProp
                         [
                             'name' => 'return',
                             'description' => '\\ArrayIterator'
+                        ],
+                        [
+                            'name' => 'phpstan-return',
+                            'description' => '\\ArrayIterator<int, '. $firstProperty->getType() .'>'
```
veewee, line 82: Not sure if it is always an int.
It might as well be a keyed array right?

Maybe `array-key` is a better alternative.

staabm: great catch, fixed.

staabm:
> > isn't this problem already resolved with the IteratorAssembler ?
> > since the PR only changes the code generated by IteratorAssembler it should not affect code, which doesn't use it?
> 
> No, see [psalm.dev/r/add3f7c288](https://psalm.dev/r/add3f7c288).

had a closer look at the example now.

the classes as beeing generated - even before this PR - will error with psalm.
I think its a different story, to make the generated classes itself green on static analysis tools.

the goal of the PR is to make static analysis tools aware of the types beeing involved in call-sites of the used objects (so the actuall application which is based on top of soap-client).

veewee:
Agree, making the class itself green is currently not possible due to the occurs limitation in ext-soap.
This PR introduces a nice enhencement already. Thanks! :)

PR #401, opened 2021-11-26
work arround differences between guzzle-psr7 and nyholm-psr7 requests

staabm:
improve PSR7 compat in case streams were already read.
~~use magic __toString() instead of getContents() because otherwise the stream might not return the actual content~~

see https://github.com/Nyholm/psr7/issues/189 for more issue details and a repro.

tbh. I am not sure whether this fix is correct or whether there is a problem in the different PSR7 implementations.

I guess we fixed similar things in the past with e.g. https://github.com/phpro/soap-client/pull/325

### DIFF
--- a/src/Phpro/SoapClient/Xml/Xml.php
+++ b/src/Phpro/SoapClient/Xml/Xml.php
@@ -90,7 +90,8 @@ class Xml
     public static function fromStream(StreamInterface $stream): Xml
     {
         $xml = new DOMDocument();
-        // use magic __toString() instead of getContents() because otherwise the stream might not return the actual content
+        // use magic __toString() instead of getContents() because otherwise the stream
+        // might not return the actual content if already read beforehand.
         $xml->loadXML($stream->__toString());
 
         /** @phpstan-ignore-next-line */
--- a/src/Phpro/SoapClient/Xml/Xml.php
+++ b/src/Phpro/SoapClient/Xml/Xml.php
@@ -90,7 +90,8 @@ class Xml
     public static function fromStream(StreamInterface $stream): Xml
     {
         $xml = new DOMDocument();
-        $xml->loadXML($stream->getContents());
+        // use magic __toString() instead of getContents() because otherwise the stream might not return the actual content
+        $xml->loadXML($stream->__toString());
 
         /** @phpstan-ignore-next-line */
         return new static($xml);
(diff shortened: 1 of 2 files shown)


src/Phpro/SoapClient/Xml/Xml.php
```diff
@@ -90,7 +90,8 @@ public function xpath(string $expression, DOMNode $contextNode = null): \DOMNode
     public static function fromStream(StreamInterface $stream): Xml
     {
         $xml = new DOMDocument();
-        $xml->loadXML($stream->getContents());
+        // use magic __toString() instead of getContents() because otherwise the stream might not return the actual content
```
staabm, line 93:
```suggestion
        // use magic __toString() instead of getContents() because otherwise the stream
        // might not return the actual content if already read beforehand.
```

veewee:
In v2, we use a combination of rewind and (string):
https://github.com/php-soap/psr18-transport/blob/main/src/Xml/Loader/Psr7StreamLoader.php

Might that be a better solution?

staabm:
casting to string with `(string)` is the same as invoking `__toString` and the psr contract defines the impl must rewind on its own in https://github.com/php-fig/http-message/blob/efd67d1dc14a7ef4fc4e518e7dee91c271d524e4/src/StreamInterface.php#L14-L28

but if you prefer for consistency with you 2.x branch I could adjust and verify the fix..

clxmstaab:
just verfied that the suggested fix works and added a unit test which fails for me without the fix

test/PhproTest/SoapClient/Unit/Xml/SoapXmlTest.php
```diff
@@ -160,6 +160,19 @@ function it_is_possible_to_create_from_a_psr7_stream()
...
     }
 
+    /**
+     * @test
+     */
+    function it_is_possible_to_create_from_a_already_read_psr7_stream()
+    {
+        $rawXml = $this->xml->saveXML();
+        $stream = new Stream(fopen('php://memory', 'rwb'));
+        $stream->write($rawXml);
```
staabm, line 170: same test as the one right before, just without explicitly calling `rewind` before passing the stream over to `fromStream`

staabm:
Ping @veewee

veewee:
Looks good, thanks!

PR #405, opened 2021-12-17
Bump version of laminas to 4.5.0

pottink:
Signed-off-by: Jelle Deneweth <jelle.deneweth@gmail.com>

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
     ],
     "require": {
         "php": "^8.0",
-        "laminas/laminas-code": "^4.4.0",
+        "laminas/laminas-code": "^4.5.0",
         "php-soap/engine": "^1.0",
         "php-soap/ext-soap-engine": "^1.0",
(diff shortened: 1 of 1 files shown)


jdreesen:
I don't think this is necessary.

pottink:
> I don't think this is necessary.

When using --prefer-lowest with composer it would still install 4.4.0 thus not compatible on PHP8.1.

jdreesen:
Are you sure it'll use `4.4.0` when running `--prefer-lowest` on PHP 8.1? 
I think composer will install the minimal working version for the PHP version you're running.

veewee:
I think @jdreesen has a valid point here ... 
If we bump the dependency, other projects need to upgrade laminas-code as well - which is still quite fresh.

Lets keep the dependency as-is for now. Nevertheless thanks for the PR : it points oit that the new version of laminas-code doesnt introduce any issues inside this repo!

PR #408, opened 2022-01-04
Use real type of iterator return value type hint

sippsolutions:
This changes

    /**
     * @return \ArrayIterator
     * @phpstan-return \ArrayIterator<array-key, \My\Own\Type>
     * @psalm-return \ArrayIterator<array-key, \My\Own\Type>
     */

to

    /**
     * @return \ArrayIterator|\My\Own\Type[]
     * @phpstan-return \ArrayIterator<array-key, \My\Own\Type>
     * @psalm-return \ArrayIterator<array-key, \My\Own\Type>
     */

After this change, when i use `foreach ($client->getArrayOfXyz()->getIterator() as $item)` my IDE knows what type `$item` is and offers suggestions of methods to call on `$item` when I type `$item->`.

### DIFF
--- a/src/Phpro/SoapClient/CodeGenerator/Assembler/IteratorAssembler.php
+++ b/src/Phpro/SoapClient/CodeGenerator/Assembler/IteratorAssembler.php
@@ -75,7 +75,7 @@ class IteratorAssembler implements AssemblerInterface
                     'tags' => [
                         [
                             'name' => 'return',
-                            'description' => '\\ArrayIterator'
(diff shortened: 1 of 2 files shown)


sippsolutions:
@veewee Would be awesome if you could take a look & also backport it to v1 as we unfortunately need to stick to PHP7.4 for the next months.

veewee:
Thanks for the PR! Looks good. 
Can you create a second PR that can be merged into 1.7.x ?

PR #415, opened 2022-01-27
Use return types from JsonSerializable/IteratorAggregate (fixes PHP 8.1 deprecations)

edudobay:
This fixes the new deprecation warnings in PHP 8.1.

> PHP Deprecated:  Return type of `<class>::jsonSerialize()` should either be compatible with `JsonSerializable::jsonSerialize(): mixed`, or the `#[\ReturnTypeWillChange]` attribute should be used to temporarily suppress the notice in `<file>` on line `<line>`

- I was unsure whether to keep the more general types from the interfaces (`mixed` and `Traversable`) or the specific types already present in the docblocks (`array` and `ArrayIterator`). I went for the latter.
- I looked up other cases of the same problem (in the `Assembler` namespace) but I couldn't find any — I'm not yet so familiar with this project.


veewee:
Looks good @edudobay! Thanks!

I also don't think there are other occurences, but if there are - feel free to create another PR.

PR #417, opened 2022-02-01
Drop default request Content-Length header, in favor of late computation

### LINKED ISSUE #416: Let final client decide content-length
Hi,

This line: https://github.com/phpro/soap-client/blob/c6e16145515fd08d5ca57339c2fcdb38189cd70b/src/Phpro/SoapClient/Soap/HttpBinding/Builder/Psr7RequestBuilder.php#L213

caused some serious debugging on our side, in a case where an auth header was added to the body at plugin level.

httplug happily ignores it: https://github.com/php-http/curl-client/blob/2ed4245a817d859dd0c1d51c7078cdb343cf5233/src/Client.php#L346

but when switching to sf/http-client it broke our call, because the endpoint did not expect it to be invalid.

Would you consider removing it, and let the final client deal with it.

In our case a request without content-length header still passes, so the workaround _eventually_ was fixing the auth plugin:

```diff
-            return $next($request->withBody($xml->toStream()));
+            return $next($request->withBody($xml->toStream())->withoutHeader('content-length'));
```

And then it started working with Symfony Psr18 client :tada: 

I will create a feature request also for Symfony, to set a default content-length either way :angel:

ro0NL:
Fixes #416


veewee:
Thanks!

Issue #425, opened 2022-05-11
PHP 8.1 return type Phpro\SoapClient\Soap\Driver\ExtSoap\AbusedClient::__doRequest


lcube45:
After implementing log events in my client factory, i'm getting noticed for a deprecated return type of AbusedClient::__doRequest.

I'm using PHP 8.1.5 and phpro/soap-client 1.7.0

Deprecated function: Return type of Phpro\SoapClient\Soap\Driver\ExtSoap\AbusedClient::__doRequest($request, $location, $action, $version, $oneWay = 0) should either be compatible with SoapClient::__doRequest(string $request, string $location, string $action, int $version, bool $oneWay = false): ?string, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice.

veewee:
Hello, 
Looks like you are doing it wrong .... v1.7 does not support 8.1.

https://github.com/phpro/soap-client/blob/60ff899a2a1609f0f5fdefea7f29303774ea1c0f/composer.json#L15

Please upgrade to 2.x, which supports php 8.1.

lcube45:
Hey,

Sorry for my mistake, composer did not give me the right version package suggestion.
I had to mitigate this with a dirty trick composer require phpro/soap-client:*.
And i didn't check the other tags, my bad.

I will switch to the correct package version considering my php requirement.

Great work and package

Cheers

lcube45:
I know now why the suggestion was bad
I'm stuck with a locked version of symfony/console (4.4.34) which doesn't meet the 2.x dependencies of your package.
Searching a way to achieve this in a Drupal 9.x project who use symfony 4.x...

PR #426, opened 2022-05-11
Fix: deprecated notice AbusedClient

lcube45:
In order to fix a deprecated notice with PHP 8.1.

Deprecated function: Return type of Phpro\SoapClient\Soap\Driver\ExtSoap\AbusedClient::__doRequest($request, $location, $action, $version, $oneWay = 0) should either be compatible with SoapClient::__doRequest(string $request, string $location, string $action, int $version, bool $oneWay = false): ?string, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice.


veewee:
See https://github.com/phpro/soap-client/issues/425#issuecomment-1123426940

PR #427, opened 2022-05-16
Feature: ConstructorAssembler add option to call parent constructor

gigerIT:
Option to generate constructors which call their parent constructor.
Adds the functionality to automatically set properties etc. to all extending classes from one single base class (parent).

### DIFF
--- a/src/Phpro/SoapClient/CodeGenerator/Assembler/ConstructorAssembler.php
+++ b/src/Phpro/SoapClient/CodeGenerator/Assembler/ConstructorAssembler.php
@@ -75,6 +75,10 @@ class ConstructorAssembler implements AssemblerInterface
             'shortdescription' => 'Constructor'
         ]);
 
+        if ($this->options->useCallParentConstructor())) {
+            $body[] = 'parent::__construct();';
+        }
+
         foreach ($type->getProperties() as $property) {
             $body[] = sprintf('$this->%1$s = $%1$s;', $property->getName());
             $withTypeHints = $this->options->useTypeHints() ? ['type' => $property->getType()] : [];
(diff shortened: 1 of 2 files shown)


veewee:
To be honest, I don't think that is a good solution.
We achieve similar results by using an event listener that injects a valid context instead.
That way, you can control from above what happens to your requests.
This makes it easier to maintain this important context and to regenerate types whenever you please.

A solution might look like this

```php
class RequestWithContext
{
    protected ContextType $Context;
    public function setContext(ContextType $context)
    {
        $this->Context = $context
    }

}

class GetRoutesRequestType extends RequestWithContext implements RequestInterface
{
    /**
     * @var string
     */
    private $OperatorCode;
}

```

```php
class AddContextListener
{
    public function onRequest(RequestEvent $event)
    {
        $request = $event->getRequest();
        if (!$request instanceof RequestWithContext) {
            return;
        }

        $context = new ContextType();
        $context->setAgentAccountNumber('ABCD123456');
        $context->setTimeStamp('2019-01-01T00:00:00');
        $context->setCountryCode('CHE');
        $context->setCurrency('EUR');
        $context->setLanguagePrefCode('en');
        $context->setTransactionId(rand(10000000, 99999999));
        $context->setUser('Test001');
        $context->setExternalBookingReference('Test001');

        $request->setContex($context);
    }
}
```

The problem with calling the parent constructor is that it might have required constructor parameters that need to be filled in. This solution simply calls it with no parameters - which will most likely result in error situations.


I'm not sure that this kind of structures is a good idea for the generated types at this moment.
Maybe you can start of by using the assembler in your own codebase and see if it proves to be usefull to you in the long run?

gigerIT:
You're right, doing it with event listeners does work as well and there wouldn't be a potential source for errors because of missing parameters. 
We'll definitely try it with this approach. I don't see any other use case then for adding the parent constructor call so i'll close the PR.
Thank you.

PR #439, opened 2022-07-06
Re-add possibility to use symfony v4

sippsolutions:
Magento (in its latest release 2.4.4) still uses `symfony/console` in Version 4.4 with PHP8.1:
`magento/framework 103.0.4 requires symfony/console (~4.4.0)`

The possibility to use Symfony v4 was removed in cc141b3b743e2f535172a259343fde2447818713 with implementing PHP8.1 compatibility, despite Symfony v4 also supports PHP8.

@veewee As we're currently upgrading our customer's project to Magento 2.4.4 and want to use the latest version of `phpro/soap-client` we'd be very happy if you could merge this asap.

Thanks in advance.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,10 +19,10 @@
         "php-soap/psr18-transport": "^1.0",
         "psr/event-dispatcher": "^1.0",
         "psr/log": "^1.0 || ^2.0 || ^3.0",
-        "symfony/console": "~5.3 || ~6.0",
-        "symfony/event-dispatcher": "~5.3 || ~6.0",
-        "symfony/filesystem": "~5.3 || ~6.0",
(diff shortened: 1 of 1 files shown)


veewee:
Sorry for the late response.

We will only support 2 major symfony version, because that's about all the pain I can bare maintaining in my spare time.
Maybe you have better luck asking the magento team to get their dependencies a bit more up to date.

You should be able to use an older version of this package in the meantime.

sippsolutions:
Unfortunately we need to use a library that depends on a pretty new version of soap-client, so we cannot use an older version of your package. I will try to escalate this to Adobe but I don't think that will lead to a solution in a timely manner ... 😃 
Very sad but I can relate - I will use a forked version of your repository in the meantime.

Issue #444, opened 2022-08-26
Allow getter return null value.


yenyasinn:
When generated class is initialised all its properties are NULL. Currently all getters have to return value of specific type, but in reality they can return NULL.

We should check whether value contains value or not first using has* method. Also It is a big issue in serialisation, when value isn't set.

Getter assembler should have ability return NULL value.

yenyasinn:
PR #445 for review is provided.

veewee:
Hello,

Please take a look at 
https://github.com/phpro/soap-client/issues/442



This issue has been reported many times already.
Ext-soap does not provide the minOccurs / maxOccurs information - meaning we can't reliably set it to nullable.
https://github.com/phpro/soap-client/blob/master/docs/known-issues/ext-soap.md#occurs

Currently, this issue can be avoided by not generating too strict types in the soap-client and optionally by using the [IteratorAssembler](https://github.com/phpro/soap-client/blob/master/docs/code-generation/assemblers.md#iteratorassembler).

This issue has been reported a few times: #346, #231, #282, #344,  ...

Until we have a better way of getting the data back from the WSDL, there is no simple solution for this.

veewee:
Hello,

We noticed this issue got reported many times and want to tackle it at its core!
Therefore, we've planned a [project that will work on better types support](https://github.com/php-soap/.github/blob/main/PROJECTS.md#better-types).
It’s going to be a huge project, so we are looking for ways to make development possible during our business hours instead of in our developers precious spare time.

In case you want this feature as badly as us: **[find out how you can support this project here 💚](https://github.com/php-soap/.github/blob/main/HELPING_OUT.md)**!

PR #445, opened 2022-08-26
#444: Allow getter return null value.

yenyasinn:


veewee:
That wouldn't be correct: Wether the type is nullable or not must be defined by minOccurs.
By making ALL properties null, your static analyzer would remind you to always do a null-check before using any value, even though minOccurs could in fact be 1 and therefor not nullable.

Not a good solution IMO.

Nothing stops you from creating your own assembler and use that one instead though.

Issue #446, opened 2022-08-31
Add check for available classes


yenyasinn:
There is todo in TypeChecker::isClassType() https://github.com/phpro/soap-client/blob/master/src/Phpro/SoapClient/CodeGenerator/Util/TypeChecker.php#L46-L50 - "add check for available classes".

The aim of this ticket to implement this check.

yenyasinn:
@veewee how did you plan to check if the class is available?

We can not use simple `class_exists()` since types that should be used in property description aren't generated yet.

veewee:
It has been a long time since that code was created.
In an old version, it indeed checked if the class exists. Until we noticed this did not work because the code was not generated yet.

I don't see an easy solution ATM, since it all are static functions.
Will have to go through some code to see where and how it is used and how it is supposed to work.
Since it is like this for about 5 years, de might as well drop that function if we don't find a better solution

yenyasinn:
PR https://github.com/phpro/soap-client/pull/447 has been opened.

I have not found a way how to check whether classes exists or not, since classes aren't existed during generation. 
Added a regular expression to check that provided string is a class name.
Actually not sure that we need `TypeChecker` at all. It is used in `FluentSetterAssembler` assembler only and actually `Normalizer` provide us already prepared types. Might be solution https://github.com/phpro/soap-client/pull/307#issuecomment-652787375 isn't so bad?

veewee:
thanks @yenyasinn,

Hmm good find!
If it is only being used in `FluentSetterAssembler`, we should probably nuke the whole `TypeChecker` class.
Can we generate the parameters like we do it in `SetterAssembler` instead?
That one indeed uses the normalizer through `\Phpro\SoapClient\CodeGenerator\Model\Property::getCodeReturnType()`

https://github.com/phpro/soap-client/blob/9c616b677df4c7038a17a5c5390becadf3f7b55e/src/Phpro/SoapClient/CodeGenerator/Assembler/SetterAssembler.php#L52-L55

This way, it works the same over all setters.

yenyasinn:
Updated PR https://github.com/phpro/soap-client/pull/447 - expanded regular expression to validate class name.

Created PR https://github.com/phpro/soap-client/pull/448 - TypeChecker has been removed completely. IMHO this solution looks better than in PR 447 with regexp.

veewee:
Closing this one - it will be dealt with in https://github.com/phpro/soap-client/issues/464

PR #447, opened 2022-09-02
#446: Add check for available classes.

yenyasinn:
See https://github.com/phpro/soap-client/issues/446

### DIFF
--- a/src/Phpro/SoapClient/CodeGenerator/Util/TypeChecker.php
+++ b/src/Phpro/SoapClient/CodeGenerator/Util/TypeChecker.php
@@ -45,8 +45,9 @@ class TypeChecker
      */
     private static function isClassType(string $type): bool
     {
-        //todo add check for available classes
-
+        if (preg_match('/^\\\[A-Z0-9]{1,}[a-zA-Z0-9\\\]+$/', $type)) {
+            return true;
+        }
         return false;
     }
 }
--- a/src/Phpro/SoapClient/CodeGenerator/Util/TypeChecker.php
+++ b/src/Phpro/SoapClient/CodeGenerator/Util/TypeChecker.php
@@ -45,7 +45,7 @@ class TypeChecker
      */
     private static function isClassType(string $type): bool
     {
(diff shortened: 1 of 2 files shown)


src/Phpro/SoapClient/CodeGenerator/Util/TypeChecker.php
```diff
@@ -45,8 +45,9 @@ private static function isInternalPhpType(string $type): bool
      */
     private static function isClassType(string $type): bool
     {
-        //todo add check for available classes
-
+        if (preg_match('/^\\\[A-Z0-9]{1,}[a-zA-Z0-9\\\]+$/', $type)) {
```
ademarco, line 48: I've tested the regex above on https://regex101.com/ and it didn't seem to work well. The suggested one below works and covers more cases:

```suggestion
        if (preg_match('/^\\?[a-zA-Z0-9_]{1,}[a-zA-Z0-9_\\]+$/', $type)) {
```

test/PhproTest/SoapClient/Unit/CodeGenerator/Util/TypeCheckerTest.php
```diff
@@ -35,6 +35,7 @@ function typeProvider()
             ['array', true],
             ['callable', true],
             ['iterable', true],
+            ['\KnownClass', true],
```
ademarco, line 38: We could add these cases here as well, so that we check the regex more thoroughly:
```
class_name
CalssName
\class_name
\CalssName
class_name\class_name
CalssName\className
\class_name\class_name
\CalssName\className\fdsaa
```

ademarco (changes requested):

src/Phpro/SoapClient/CodeGenerator/Util/TypeChecker.php
yenyasinn: I've updated regular expression - lower case characters in the beginning of class and underscore are allowed. But I left check that class name has to be started with `\`

veewee:
Closing this one in favour for #448

PR #406, opened 2021-12-17
Don't import known types

janvernieuwe:
Fixes #377

PR #448, opened 2022-09-02
#446: Do not use TypeChecker in FluentSetterAssembler.

yenyasinn:
See https://github.com/phpro/soap-client/issues/446

### DIFF
--- a/src/Phpro/SoapClient/CodeGenerator/Assembler/FluentSetterAssembler.php
+++ b/src/Phpro/SoapClient/CodeGenerator/Assembler/FluentSetterAssembler.php
@@ -4,9 +4,7 @@ namespace Phpro\SoapClient\CodeGenerator\Assembler;
 
 use Phpro\SoapClient\CodeGenerator\Context\ContextInterface;
 use Phpro\SoapClient\CodeGenerator\Context\PropertyContext;
-use Phpro\SoapClient\CodeGenerator\Model\Property;
 use Phpro\SoapClient\CodeGenerator\Util\Normalizer;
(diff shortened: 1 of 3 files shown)


veewee:
Thanks for the PR.
It looks like there is still an occurence somewhere?

yenyasinn:
Yes, it is used in ClientMethodAssembler

yenyasinn:
No, please have a look https://github.com/phpro/soap-client/blob/master/src/Phpro/SoapClient/CodeGenerator/Assembler/ClientMethodAssembler.php#L217. 
It some another case.

veewee:
@janvernieuwe Do you remember why that check was added to the `ClientMethodAssembler`?
The linked ticket in #406 doesn't seem related to the fix?

It seems like we can drop the `TypeChecker` if we change the logic inside the `ClientMethodAssembler`.

I was thinking about changing


https://github.com/phpro/soap-client/blob/9c616b677df4c7038a17a5c5390becadf3f7b55e/src/Phpro/SoapClient/CodeGenerator/Model/ClientMethod.php#L88-L91

To make it work more like 

https://github.com/phpro/soap-client/blob/9c616b677df4c7038a17a5c5390becadf3f7b55e/src/Phpro/SoapClient/CodeGenerator/Model/Property.php#L64-L71

That way, we could check for the trailing `\` or if the fqcn contains a `\` to return the skip.
Internally it will behave smarter, since it will use both the list of reserved keywords and Type suffixer.

What do you think?

janvernieuwe:
@veewee It is related to the fix, if i recall correctly it tried to use root (\) for internal types like string.
Your suggested fix looks good to me, a test was also added to check for regression on this so we should be good.

PR #456, opened 2023-01-24
Improve ValidatorSubscriber error messages by including property path.

ademarco:
This PR aims at improving the error messages thrown by ValidatorSubscriber, by including the property path on which the error message is occurring.

For example, at the moment all we get is:

```
Phpro\SoapClient\Exception\RequestException: This value should not be null.
```

With this PR, we would know which property to check:

```
Phpro\SoapClient\Exception\RequestException: requestDetails.products: This value should not be null.
```


veewee:
Looks good, thanks!

PR #468, opened 2023-05-24
Generated constructor is descripbed by @param instead of @var.

h4kuna:
The phpstan does not understand annotation `@var` for parameters in constructor for generated files.

You are using @param too in [your code](https://github.com/phpro/soap-client/blob/439f92226e4c93a0663d843dcaf65d62b16de54a/src/Phpro/SoapClient/Type/MixedResult.php#L20).


veewee:
Thanks for the PR.

PR #478, opened 2023-07-27
feat: add common naming convention option for getter assmebler

emmanuel-averty:
This PR propose to add an option in GetterAssemblerOptions (`commonNamingConvention`) in order to have getters that the popular `symfony/property-access` component could use.

If property name contains an underscore (i.e. `my_prop`), the getter generated without this option is `getMy_prop`. Whereas `symfony/property-access` waits for `getMyProp`. (https://symfony.com/doc/current/components/property_access.html#using-getters)

This options allows to use Symfony Serializer and Symfony tools like `dd`.

### DIFF
--- a/spec/Phpro/SoapClient/CodeGenerator/Util/NormalizerSpec.php
+++ b/spec/Phpro/SoapClient/CodeGenerator/Util/NormalizerSpec.php
@@ -72,6 +72,17 @@ class NormalizerSpec extends ObjectBehavior
         $this->normalizeProperty('My-./final*prop_123')->shouldReturn('MyFinalProp_123');
     }
 
+    function it_noramizes_properties_for_common_naming_convention()
+    {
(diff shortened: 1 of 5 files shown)


veewee:
Thanks for the PR.

I'm wondering if it make sense to make this configurable or not:

Since we are making some big code generation improvements on [v3](https://github.com/phpro/soap-client/tree/v3.x), it might be a better idea to make this the default behaviour instead.
If one really want to make this configurable, it might make more sense to introduce a `case` enum that could be set to `snake`, `camel`, ....

Do note that this should work similar on all possible assemblers: getters, setters, fluent setters, immutable setters, ....

What do you think about this?

emmanuel-averty:
I agree with you. I think this should be the default behavior. I didn't want to introduce BC breaks in my proposal.

veewee:
Fixed in #484 (and #486)

PR #503, opened 2024-03-05
Normalize unsignedint

RosskoDCA:
UnsignedInt type is generated as custom type and later it fails when trying assign integer value to object. The unsignedInt is native soap type which means there is not type definition in WSDL.

For now the best solution is map this value to float.

If you have a better idea, feel free to suggest it, I'll try my best to implement it.


RosskoDCA:
unfortunatelly, my bad, this will not work as i expected :)

veewee:
Feel free to open up an issue with some more detailed information instead if this is still a problem.

PR #516, opened 2024-05-16
Consider nullability for properties

### LINKED ISSUE #515: Set null as a default value for nullable properties
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes/no
| Version     | 3.20

#### Summary

PropertyAssembler should assign null as a default value for nullable properties.

#### Current behavior

PropertyAssembler can be configured to assume that all properties are nullable but it will not correctly check for nullability and only then assign null as a default.

#### How to reproduce

Use `PropertyAssembler` as is with a nullable property and the generated code will be `private ?Type $prop;`.

#### Expected behavior

Use `PropertyAssembler` as is with a nullable property and the generated code will be `private ?Type $prop = null;` to prevent "uninitialized property" errors.

lstrojny:
Fixes #515


veewee:
Thanks!

PR #518, opened 2024-05-17
Fix invalid array bounds during static analysis

### LINKED ISSUE #26: Incorrect array bounds in docs
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | ?
| Version     | 0.10.0

#### Summary

The array key boundaries generated in XsdTypeFormatter are incorrect. They attempt to define the length of the array but instead they define the shape of the keys.

#### Current behaviour

Given this element:

```xml
<xs:element name="SomeElement" type="xs:string" maxOccurs="20">
```

The resulting type is this:

```
array<int<1,20>,string>
```

This would reject a 0 indexed array like this (because it only allows keys from 1 to 20):

```
['foo']
```

Such an array would be totally fine though.

#### Expected behaviour

Instead of limiting the array keys I think it would be good enough to differentiate list and non-empty-list.

```
$isList ? $min > 0 ? 'non-empty-list' : 'list' : '',
```

See https://github.com/php-soap/wsdl-reader/blob/32d32e01bb3347337b0c52ebff2024ed34bdae1a/src/Formatter/XsdTypeFormatter.php#L21

veewee:
Fixes https://github.com/php-soap/wsdl-reader/issues/26

@lstrojny : I decided to go this way since it is the easiest way to go forward.

The array-bounds will always be calculated based on int<0, maxOccurs - 1>:
When minOccurs > 0, the array will become a non-empty-array.

Examples:

```php
// minOccurs=0, maxOccurs=-1
array<int<0, max>, Tv>

// minOccurs=1, maxOccurs=-1
non-empty-array<int<0, max>, Tv>

// minOccurs=0, maxOccurs=0
array<never, Tv>

// minOccurs=0, maxOccurs=1
array<int<0,0>, Tv>

// minOccurs=0, maxOccurs=2
array<int<0,1>, Tv>

// minOccurs=1, maxOccurs=2
non-empty-array<int<0,1>, Tv>
```

This should solve your issue.

If someone cares enough to improve this, I'm always open into considering something like:

```php
array{0: string, 1 ?: string, ...string}
```

(limited to x amount of items to avoid too large shapes being generated)

Yet I think this should be sufficient for now.

### DIFF
--- a/src/Phpro/SoapClient/CodeGenerator/TypeEnhancer/Calculator/ArrayBoundsCalculator.php
+++ b/src/Phpro/SoapClient/CodeGenerator/TypeEnhancer/Calculator/ArrayBoundsCalculator.php
@@ -7,16 +7,22 @@ use Soap\Engine\Metadata\Model\TypeMeta;
 
 final class ArrayBoundsCalculator
 {
+    /**
+     * All lists will start from index 0.
+     *
+     * The maximum amount of items in the list will be maxOccurs - 1, since arrays are zero-index based.
+     * Edge cases like -1 (unbounded) and 0 (empty) are handled as well.
+     *
+     * These bounds don't take into account minOccurs, since minOccurs still starts from 0.
+     */
     public function __invoke(TypeMeta $meta): string
     {
-        $min = $meta->minOccurs()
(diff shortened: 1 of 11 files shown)


lstrojny:
Tried and the solutions works perfectly fine. Thank you!

Issue #531, opened 2024-07-21
Can't instantiate custom type for GUIDs


akayumeru:
|    Q        |   A
|------------ | -----
| Version     | 4.0.0

### Support Question

Hello. I have some problem with custom data types. I created a type encoder and also added type replacer. However, the generator gives me this type as a converted name in my namespace. I am using ramsey's uuid library and wanna use UuidInterface for GUIDs. Here's a code:

```php
class GuidEncoder implements XmlEncoder
{
    public function iso(Context $context): Iso
    {
        return new Iso(
            fn (UuidInterface $value): string => $value->toString(),
            static fn (string $value): UuidInterface => Guid::fromString($value)
        );
    }
}
```

```php
class GuidTypeReplacer implements TypeReplacer
{
    public function __invoke(XsdType $xsdType): XsdType
    {
        $check = new IsOfType('http://microsoft.com/wsdl/types/', 'guid');

        if (!$check($xsdType)) {
            return $xsdType;
        }

        return $xsdType->copy('guid')->withBaseType(UuidInterface::class);
    }
}
```

In code generation config:
```php
->setTypeReplacementStrategy(
    TypeReplacers::defaults()
        ->add(new GuidTypeReplacer())
)
```

And in ClientFactory:
```php
->addComplexTypeConverter(
    'http://microsoft.com/wsdl/types/',
    'guid',
    new GuidEncoder()
)
```

So, the result of code generation is:
```php
/**
 * @var \LazyTechwork\MyClient\Types\RamseyUuidUuidInterface
 */
private \LazyTechwork\MyClient\Types\RamseyUuidUuidInterface $ID;
```

instead of:
```php
/**
 * @var \Ramsey\Uuid\UuidInterface
 */
private \Ramsey\Uuid\UuidInterface $ID;
```

Can you help me deal with this problem?

veewee:
Hello @LazyTechwork,

Thanks for trying it v4 out and reporting this issue.

At the moment, Adding types from a different PHP namespace than the once where the types are generated (or PHP's built-in types) is not supported yet.
I'll have to take a look at how it can be implemented.

In the meantime, you could add a smalle `Guid` wrapper-class in the folder where the types are generated. Something like:

```php
class Guid 
{
     private UuidInterface $id;

    // And also fromString and toString
}
```

That way you can use it as you'dd expect.


Given the XML contains a simple XML string, you'dd probably need to change the ClientFactory to use a simple type instead of a complex type:

```diff
[...]

PR #548, opened 2024-09-27
Use external classes as property types.

### LINKED ISSUE #543: v4 codegen: php type encoder
|    Q        |   A
|------------ | -----
| Version     | 4.0.0-alpha1

### Support Question

I've been trying to generate soap client type definitions with the built-in generator, but there seems to be no way to specify e.g `\Cake\Chronos\Chronos` instead of `\DateTimeImmutable`. (`Phpro\SoapClient\CodeGenerator\Util\Normalizer` is set to normalize simple types to hard-coded interfaces)

I can specify custom `SimpleType` encoders for the client, but not for generating a client. :)

veewee:
Fixes #543, #531

Allows to specify a PHP class as property type:

```php
use Phpro\SoapClient\Soap\Metadata\Manipulators\TypeReplacer\TypeReplacer;
use Soap\Engine\Metadata\Model\XsdType;
use Soap\WsdlReader\Metadata\Predicate\IsOfType;

class GuidTypeReplacer implements TypeReplacer
{
    public function __invoke(XsdType $xsdType): XsdType
    {
        $check = new IsOfType('http://microsoft.com/wsdl/types/', 'guid');
        if (!$check($xsdType)) {
            return $xsdType;
        }

        return $xsdType->copy(\Ramsey\Uuid\UuidInterface::class);
    }
}
```

This will result in:

```php

namespace MyNamespace;

class MyType
{
    /**
     * Type specific docs
     *
     * @var \Ramsey\Uuid\UuidInterface
     */
    private \Ramsey\Uuid\UuidInterface $prop1;
}
```

(it will also be known this way in the other assemblers like Getters, Setters, ...)

/cc @rauanmayemir: The implementation is a bit tricky, but it does the job. Feel free to play around with it.

### DIFF
--- a/docs/drivers/metadata.md
+++ b/docs/drivers/metadata.md
@@ -49,7 +49,7 @@ return Config::create()
 
 ### Type replacements
 
-Depending on what XML encoders you configure, you might want to replace some types with other types.
+Depending on what [SOAP encoders](https://github.com/php-soap/encoding#encoder) you configure, you might want to replace some types with other types.
 Take following example:
 
 By default, a "date" type from the XSD namespace `http://www.w3.org/2001/XMLSchema` will be converted to a `DateTimeImmutable` object.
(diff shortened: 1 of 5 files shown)


rauanmayemir:
Could you rebase with changes from main?

veewee:
@rauanmayemir done

rauanmayemir:
I believe this works. 🙌 
I think all these changes warrant 4.0.0-alpha2, no?

veewee:
Yes, tagged it
https://github.com/phpro/soap-client/releases/tag/4.0.0-alpha2

PR #552, opened 2024-10-22
Rework code generator models for easier 3rd-party class type encoders

### LINKED ISSUE #543: v4 codegen: php type encoder

veewee:
Fixes https://github.com/phpro/soap-client/issues/543

@rauanmayemir  This should fix the `Decimal` type replacement issues you encountered during code generation. Can you validate if this works for you?


rauanmayemir:
@veewee this works. 👌

PR #562, opened 2025-01-09
Mark non-intersecting types as nullable

veewee:
This PR marks non-intersecting types as null whilst using the IntersectDuplicateTypesStrategy.

This results in nullable properties (getters, ...) during code generation with a default value of `null`.
So when the item is not available and you do try to access the property, you won't get an uninitialized error but null instead.

For example
```php
    /**
     * @var null | string
     */
    private ?string $Time = null;

    /**
     * @return null | string
     */
    public function getTime() : ?string
    {
        return $this->Time;
    }
```

This means that your code needs to validate wether the value is present or not (at all times) before using non intersecting properties.

### DIFF
--- a/src/Phpro/SoapClient/Soap/Metadata/Manipulators/DuplicateTypes/IntersectDuplicateTypesStrategy.php
+++ b/src/Phpro/SoapClient/Soap/Metadata/Manipulators/DuplicateTypes/IntersectDuplicateTypesStrategy.php
@@ -10,7 +10,15 @@ use Soap\Engine\Metadata\Collection\PropertyCollection;
 use Soap\Engine\Metadata\Collection\TypeCollection;
 use Soap\Engine\Metadata\Model\Property;
 use Soap\Engine\Metadata\Model\Type;
+use Soap\Engine\Metadata\Model\TypeMeta;
+use function Psl\Iter\contains;
+use function Psl\Iter\first;
+use function Psl\Iter\reduce;
+use function Psl\Type\instance_of;
 use function Psl\Type\non_empty_string;
+use function Psl\Vec\flat_map;
+use function Psl\Vec\map;
+use function Psl\Vec\values;
 
 final class IntersectDuplicateTypesStrategy implements TypesManipulatorInterface
 {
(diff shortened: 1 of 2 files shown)


veewee:
@rauanmayemir 

I'm moving my service to v4 and thought this might be a sensible change.
Does it impact your services?

rauanmayemir:
Does it just set default `null` value for nullable types?

I'm not sure if it's just my generator config, but I already have that in place, nullable types are already marked e.g like this:
```php
    /**
     * Document number
     *
     * @var null | string
     */
    protected ?string $DocumentNum = null;

    /**
     * @return null | string
     */
    public function getDocumentNum() : ?string
    {
        return $this->DocumentNum;
    }

    /**
     * @param null | string $DocumentNum
     * @return $this
     */
    public function setDocumentNum(?string $DocumentNum) : static
    {
        $this->DocumentNum = $DocumentNum;
        return $this;
    }
```

veewee:
@rauanmayemir  it depends on how you configured code generation.
For nullable types, the code generation automatically sets this `null` default value.
However, you can control to do this for all properties by specifying `PropertyAssemblerOptions::optionalValue`


This PR is about when there are duplicate type-names and when you choose to intersect those types.
For example:

DuplicateType:
- **a**
- b

DuplicateType:
- **a**
- c

In this case the intersection will:

DuplicateType:
- a
- b -> always nullable
- c-> always nullable

I think your answer answered my question as well. :)

rauanmayemir:
Indeed, I played with options and see that deliberately left out the optional value:
```php
                    new Rules\AssembleRule(
                        new Assembler\PropertyAssembler(
                            Assembler\PropertyAssemblerOptions::create()
                                ->withVisibility(PropertyGenerator::VISIBILITY_PROTECTED)
//                                ->withOptionalValue()
                            ,
                        )
                    ),
```

It's odd that the default value is still applied.

rauanmayemir:
Anyway, I can always tweak the generator and change the behavior if I need it to be different from default.

veewee:
> It's odd that the default value is still applied.

There is a overwrite during code generation for nullable types.
When a type is nullable, we always set the default to `null` (to avoid non initialized errors)

PR #580, opened 2025-06-13
php 8.4 support

kevujin:
version 3.1 of azjezz/psl has some very minor changes along with increased version of php from 8.3 to 8.4


veewee:
Current range already supports the complete 3.x range of psl.

PR #595, opened 2025-11-04
Update ClientConstructorAssembler to use type instead of phpdoc

ayrtonandino:
This pr changes the output generated by `ClientConstructorAssembler` from:

```php
use Phpro\SoapClient\Caller\Caller;
    
class MyClient
{
    /**
     * @var Caller
     */
    private $caller;

    public function __construct(\Phpro\SoapClient\Caller\Caller $caller)
    {
```

to:

```php
class MyClient
{
    private \Phpro\SoapClient\Caller\Caller $caller;

    public function __construct(\Phpro\SoapClient\Caller\Caller $caller)
    {
```

Also upgrade `laminas/laminas-code` to latest version since GrumPHP didn't let me commit due to changes in [laminas/laminas-code/pull/208](https://github.com/laminas/laminas-code/pull/208) 

### why this change

I'm using phpstan with `tomasvotruba/type-coverage` plugin, and the only missing type in my codebase is this `Caller` interface, it works the same and the type is already validated in the constructor, so It shouldn't break anything.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,14 +31,14 @@
         "guzzlehttp/guzzle": "^7.5.0",
         "nyholm/psr7": "^1.5",
         "php-http/vcr-plugin": "^1.2",
-        "php-parallel-lint/php-parallel-lint": "^1.3",
+        "php-parallel-lint/php-parallel-lint": "^1.4",
         "phpro/grumphp-shim": "^2.17",
-        "phpspec/phpspec": "~7.2 || ~8.0",
+        "phpspec/phpspec": "^8.0",
         "phpspec/prophecy-phpunit": "^2.0.1",
         "phpstan/phpstan": "^2.1",
-        "phpunit/phpunit": "~12.4",
+        "phpunit/phpunit": "^12.4",
         "squizlabs/php_codesniffer": "^4.0",
-        "symfony/cache": "^6.4 || ^7.0"
+        "symfony/cache": "^7.0"
     },
     "config": {
         "sort-packages": true,
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
     "require": {
         "php": "~8.3.0 || ~8.4.0 || ~8.5.0",
         "azjezz/psl": "^3.0 || ^4.0",
-        "laminas/laminas-code": "^4.14.0",
+        "laminas/laminas-code": "^4.17.0",
         "php-soap/cached-engine": "~0.3",
         "php-soap/engine": "^2.14.0",
         "php-soap/encoding": "~0.15",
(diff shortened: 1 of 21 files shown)


src/Phpro/SoapClient/CodeGenerator/Assembler/ClientConstructorAssembler.php
```diff
@@ -32,20 +30,14 @@ public function assemble(ContextInterface $context)
 
         $class = $context->getClass();
         try {
-            $caller = $this->generateClassNameAndAddImport(Caller::class, $class);
```
veewee, line 35: I don't think this line should be removed.
The class should be imported. However, instead of using `Called::class` as type directly, we need to use `$caller`

composer.json
```diff
@@ -33,7 +33,7 @@
         "php-http/vcr-plugin": "^1.2",
         "php-parallel-lint/php-parallel-lint": "^1.3",
         "phpro/grumphp-shim": "^2.17",
-        "phpspec/phpspec": "~7.2",
+        "phpspec/phpspec": "~7.2 || ~8.0",
```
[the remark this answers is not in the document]
veewee: Could we just bump to 8 alltogether? It's a dev dependency, so we dont really need v7 as long as its supported by all php versions in our matrix.

veewee:
Thanks @ayrtonandino. Nice addition!
I've added some small remarks.

src/Phpro/SoapClient/CodeGenerator/Assembler/ClientConstructorAssembler.php
ayrtonandino: tried that, it certainly looks better doing `Caller $caller`, but `TypeGenerator::fromTypeString`, always prepends `\`, so it generates this instead, which is not valid:

```php
use Phpro\SoapClient\Caller\Caller;

class MyClient
{
    private \Caller $caller;

    public function __construct(\Caller $caller)
    {
```

thats why the old version didn't use the imported type in the constructor

veewee: Ah yes, I see. Ok to do it like this then.

Thanks for clarifying!

veewee:
Thanks!

src/Phpro/SoapClient/CodeGenerator/Assembler/ClientConstructorAssembler.php
```diff
@@ -1,19 +1,17 @@
 <?php
+
```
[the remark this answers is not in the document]
ayrtonandino: don't think this is valid in this case, from [php-fig.org](https://www.php-fig.org/psr/psr-12/) :

> When wishing to declare strict types in files **containing markup outside PHP opening and closing tags**, the declaration MUST be on the first line of the file and include an opening PHP tag, the strict types declaration and closing tag.

in that case it should be:

```php
<?php declare(strict_types=1) ?>
```

PR #610, opened 2026-05-27
GenConfigCommand - print wsdl loading exceptions when output is verbose

YasserB94:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

## Context
I spent some time debugging a local WSDL file  recently.  The path was incorrect but the exception was silently ignored, making it tedious to figure out what went wrong. The exception message pointed me straight to the issue.

I was using 5.x - I noticed the path issue I had was resolved in https://github.com/php-soap/wsdl/pull/35. _(thank you)_

## Changes
- Catch any thrown exception and output debug info on verbose runs during WSDL Loading in the GenerateConfigCommand
- `-v`  shows the exception class and message
- `-vv` also shows the stack trace

_Sorry for the formatting changes, without them I was unable to pass all pre-commit checks._

### DIFF
--- a/src/Phpro/SoapClient/Console/Command/GenerateConfigCommand.php
+++ b/src/Phpro/SoapClient/Console/Command/GenerateConfigCommand.php
@@ -2,7 +2,6 @@
 
 namespace Phpro\SoapClient\Console\Command;
 
-use Laminas\Code\Generator\FileGenerator;
 use Phpro\SoapClient\CodeGenerator\Config\ClassMapConfig;
 use Phpro\SoapClient\CodeGenerator\Config\ClientConfig;
 use Phpro\SoapClient\CodeGenerator\Config\Destination;
@@ -19,6 +18,7 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
+use Laminas\Code\Generator\FileGenerator;
 
 class GenerateConfigCommand extends Command
 {
@@ -45,13 +45,13 @@ class GenerateConfigCommand extends Command
 
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
-        $context = new ConfigContext;
+        $context = new ConfigContext();
         $io = new SymfonyStyle($input, $output);
-        $required = new NotBlankValidator;
+        $required = new NotBlankValidator();
 
         // Ask for config location:
         $destination = $input->getOption('config');
-        if (! $destination) {
+        if (!$destination) {
             $destination = $io->ask(
                 'config location (Where to put the config, including .php)',
                 'config/soap-client.php'
@@ -62,6 +62,7 @@ class GenerateConfigCommand extends Command
         $context->setWsdl($wsdlUri);
 
         $io->warning('Attempting to load WSDL... (this might take a while)');
+
         try {
             $wsdl = $this->loadWsdl($wsdlUri);
         } catch (\Throwable $e) {
@@ -77,6 +78,7 @@ class GenerateConfigCommand extends Command
             $wsdl = null;
         }
 
+
         $context->setDetectedXmlNamespaces($wsdl?->namespaces->namespaceToNameMap ?? []);
         $context->setGenerateDocblocks($io->confirm('Should methods be generated with docblocks?', true));
         $name = $io->ask(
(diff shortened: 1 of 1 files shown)


src/Phpro/SoapClient/Console/Command/GenerateConfigCommand.php
```diff
@@ -62,11 +62,19 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
 
         $io->warning('Attempting to load WSDL... (this might take a while)');
-        $wsdl = $this->loadWsdl($wsdlUri);
-
-        if (!$wsdl) {
+        try {
+            $wsdl = $this->loadWsdl($wsdlUri);
+        } catch (\Throwable $e) {
             $io->warning('Could not load the provided WSDL with default engine options.');
+            if ($output->isVerbose()) {
```
veewee: Maybe better to display this always? I don't think people will run this command with -v very often.

src/Phpro/SoapClient/Console/Command/GenerateConfigCommand.php
```diff
@@ -80,24 +88,24 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         $namespace = Normalizer::normalizeNamespace($io->ask('Namespace for your client', null, $required));
 
         // Create configuration objects
-        $typeDestination = new Destination($baseDir . DIRECTORY_SEPARATOR . 'Type', $namespace . '\\Type');
+        $typeDestination = new Destination($baseDir.DIRECTORY_SEPARATOR.'Type', $namespace.'\\Type');
```
veewee: Not sure why you needed to apply all these style changes, because this is not in line with the style in the other classes. (same for pretty much all style changes you did in this file)
What failed exactly?

src/Phpro/SoapClient/Console/Command/GenerateConfigCommand.php
```diff
@@ -62,11 +62,19 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
-        $wsdl = $this->loadWsdl($wsdlUri);
-
-        if (!$wsdl) {
+        try {
+            $wsdl = $this->loadWsdl($wsdlUri);
+        } catch (\Throwable $e) {
             $io->warning('Could not load the provided WSDL with default engine options.');
+            if ($output->isVerbose()) {
+                $io->text('<fg=red>'.$e::class.'</>');
+                $io->text($e->getMessage());
```
veewee: maybe also better to put in red.
Isn't there a helper in $io to style the error message?

veewee (commented):
Thanks for the PR.
I added some small remarks.

src/Phpro/SoapClient/Console/Command/GenerateConfigCommand.php
YasserB94: I am sorry about the formatting changes. I think a Laravel pint auto command slipped through.
I cleaned this up in [4287b26](https://github.com/phpro/soap-client/pull/610/commits/4287b266b4ef747afad53512fe7d107288ff0f5b) and used the included phpcbf instead.

src/Phpro/SoapClient/Console/Command/GenerateConfigCommand.php
YasserB94: Yes, It felt a bit aggressive since its a big red block, this is just some red text.

Though you make a good point, and it is probably more inline with the other output (Such as warning)

src/Phpro/SoapClient/Console/Command/GenerateConfigCommand.php
YasserB94: I am happy to change this. Although I have a couple of arguments against showing this at all times

For me the story went:
- Something went wrong. I got a warning, fine. But I want to know more
- Ran --help -> Saw verbose output
- Re-ran verbose and expected extra info on what went wrong

***
- Some exceptions come from deep down, dependencies,... and can be quite cryptic. The already in place warning looks more user friendly _(For example, bad xml formats would print something like Unexpected character `<;`  in a big red block - when unexpected this seems like bad UX to be)_. 
- I feel like using -v and -vv is somewhat in line with the --help message (From Symfony Console)
```sh
# /bin/soap-client wizard --help
Options:
  -v|vv|vvv, --verbose  Increase the verbosity of messages: 1 for normal output, 2 for more verbose output and 3 for debug
```
***
**TLDR:** I think it's a bit aggressive UX to always display these. But I am happy to make this change.

veewee: The problem is that: even though the config can be generated, you won't be able to use that generated config given that the loading of WSDL will fail in the next command you run (especially if you use the wizard).

If the issue is e.g. at connection level rather than on XML level, this will give you quite an idea already to start with. 
Maybe we should display the message less aggressively since it's not blocking (at least for this command) - but always display some more details than be vague about what error happened exactly?

YasserB94: Completely makes sense. I felt it would be too aggressive, but you made very good points
I have updated this in [1153321](https://github.com/phpro/soap-client/pull/610/commits/1153321f76793e3d938ba1852d808fb7594618a2)

### Summary
- Use error output for the class/message
- Pass arrays to the Symfony IO so it can handle multi line formatting for us. _Keeping larger exception messages and the stack trace nicely formatted_

**Note:** Would you like for me to squash/rebase this branch? (get rid of the formatting commits etc?)

